PR #4, opened 2018-08-01
Enable security stuff in Open API definition

xalioth:
Allows to specify a swagger securitySchemes in the top level API doc
Also allows to specify the security scheme for a whole Blueprint at once rather than specifying the security='xxx' for each functions

### DIFF
--- a/docs/openapi.rst
+++ b/docs/openapi.rst
@@ -33,6 +33,31 @@ specified as Flask application parameters:
 
    Default: ``'2.0'``
 
+Documenting Top-level Components
+------------------------------------------------------
+
+To add objects within the top-level components object, specify them in the
+API_SPEC_OPTIONS Flask application parameter:
+
+.. describe:: API_SPEC_OPTIONS
+
+   A dict to include in the top level components object of the Open API
+   definition. It is copied verbatim in the documentation.
+
+Here is an example on how to includes a Security Scheme Object:
+
+.. code-block:: python
+
+    app.config['API_SPEC_OPTIONS'] = {
+        'securityDefinitions': {
+            'MyToken': {
+                'type': 'apiKey',
+                'in': 'header',
+                'name': 'Authorization'
+            }
+        }
+    }
+
 Adding Documentation Information to the View Functions
 ------------------------------------------------------
 
@@ -97,15 +122,15 @@ for instance if it is provided by another Flask extension.)
 
 .. code-block:: python
 
-    # UUIDConverter is already known to the flask app as it is imported
+    # UUIDConverter is already known to the flask app as it is imported
     # from an extension that registers it.
     api.register_converter(UUIDConverter, 'string', 'UUID')
 
     @blp.route('/pets/{uuid:pet_id}')
         ...
 
-    # CustomConverter is defined in our application and must be registered
-    # in the Flask app. A name must be passed.
+    # CustomConverter is defined in our application and must be registered
+    # in the Flask app. A name must be passed.
     api.register_converter(CustomConverter, 'string', 'Custom',
                            name='custom')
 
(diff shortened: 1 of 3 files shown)


lafrech:
Ideally, the registration of the security definition would be done the same way definitions are registered. See this issue: https://github.com/marshmallow-code/apispec/issues/245. It shouldn't be too much work, but until it is done, or to allow more flexibility, being able to pass a dictionary makes sense.

We could also add a `spec_kwargs` argument to `Api` to pass kwargs to the `APISpec` instance, like it currently does with `spec_plugins`, for top-level options that don't need to be an application parameter but are hardcoded in the application.

Regarding the security argument to `Blueprint`, I'd like to think twice before doing this. Do you think it could be done application-wise (excluding the blueprint that serves the spec, of course)? Or would it be too much of a constraint and a blueprint granularity is needed?

And would it make sense to make it more generic, like adding a `view_docs` parameter to `Blueprint` (TODO: find a better name), so that any keyword can be specified, not only `security`:

```py
    def __init__(self, *args, **kwargs):
        self.description = kwargs.pop('description', '')
        self.view_docs = kwargs.pop('view_docs', {})

        [...]
 
        def store_method_docs(method, function):
            [...]
            endpoint_doc[method_l].update(self.view_docs)
```

xalioth:
Yes at first I wanted to add a decorator to add a security definition, but the APISpec code doesn't seem to allow adding that after initialization, i.e. it has to be passed to the APISpec constructor.

I then also implemented a solution as you suggested (adding arguments to Api constructor), but at the end it was not so clean because the args had to be passed from functions to functions internally, so I prefered the config solution, which is somewhat similar to what you did for passing the versions.

For the blueprint, I agree it could be made more generic, maybe rename `view_docs` to `default_doc` or fallback_doc ?

xalioth:
In any case, I am completely fine with another solution as long as I got a way to define my security schemes. I need it to easily test my API directly from swaggerdoc

lafrech:
> I then also implemented a solution as you suggested (adding arguments to Api constructor), but at the end it was not so clean because the args had to be passed from functions to functions internally, so I prefered the config solution, which is somewhat similar to what you did for passing the versions.

Both ways have their use case. I think using an app parameter makes sense if it is really something that may be set for a specific application instance (like URL, local path, or various settings). Hardcoded parameters are for stuff that is not meant to be changed when deploying, especially if it is made of complicated python objects you wouldn't put in a config file, or if it needs a processing (for instance, parameters, responses, errors, that you'd like to parse using MarshmallowPlugin) before being passed.

We can have both.

Yes, `default_doc` seems fine. Let's investigate if we can do it application-wise. And maybe think of an opt-out if an exception is needed (passing `{'attribute': `None`}`, for instance, could cancel a default locally). This mechanism, or a similar one, would be useful to specify error responses globally.

I won't be able to merge, let alone release, any of this before one or two weeks. In the meantime, I think you can safely use `API_SPEC_OPTIONS` with your patch, as ultimately, it should be released.

xalioth:
OK, fine for me, I'm not in a hurry, I have mu own branch. Thanks Jérôme!

lafrech:
Hi @xalioth. I just released a 0.8.0 version implementing `API_SPEC_OPTIONS` and an equivalent `spec_kwargs` parameter. Thanks for the suggestion.

I didn't take much time to think about `default_doc` but feel free to open a dedicated issue / PR if you want to discuss / implement such a feature. I have no objection. Just want to get it right before releasing.

lafrech:
> Also allows to specify the security scheme for a whole Blueprint at once rather than specifying the security='xxx' for each functions

If suppose you're using a decorator to enforce security. Can't you change it to auto-document the security scheme applied to the route? (See https://github.com/Nobatek/flask-rest-api/issues/36 for a similar issue.)

PR #13, opened 2018-10-10
Allow OPENAPI_URL_PREFIX to be '/'

fbergroth:
If ` OPENAPI_URL_PREFIX == '/'`, `_add_leading_slash` is being called with an empty string and crashes with an IndexError.

### DIFF
--- a/flask_rest_api/spec/__init__.py
+++ b/flask_rest_api/spec/__init__.py
@@ -66,7 +66,7 @@ class APISpec(apispec.APISpec):
 
 def _add_leading_slash(string):
     """Add leading slash to a string if there is None"""
-    return string if string[0] == '/' else '/' + string
+    return string if string.startswith('/') else '/' + string
 
 
 class DocBlueprintMixin:


lafrech:
Nice catch.

I didn't think of putting the docs at the root of the application, but thinking of it, I find it better and I might do the same in my apps in the future. If `OPENAPI_REDOC_PATH` is also `/` (or `''`), the landing page of the app will be the documentation page, which is nice.

I did a few more modifications in bfcda78 to allow this.

Thanks for the suggestion.

fbergroth:
@lafrech alright, cool. I'm just started out with this library, so I'll be blunt and hijack this PR to ask a question. Is there any simple way to make a mixed type? I'm trying to make a custom field that renders this type:

```
oneOf:
  - type: string
  - type: array
    items:
      type: string
```

lafrech:
You can bake a custom marshmallow field that would do that.

Regarding the documentation, I don't know. Polymorphism is still a complicated thing: https://github.com/marshmallow-code/apispec/pull/182.

I think if we implemented the `FIELD_TO_PROPERTY` mapping I'm suggesting in https://github.com/marshmallow-code/apispec/issues/172#issuecomment-415065055, you could have a workaround. Maybe a bit too manual, but better than nothing.

fbergroth:
Thanks for linking these issues.

> I think if we implemented the FIELD_TO_PROPERTY mapping I'm suggesting in marshmallow-code/apispec#172 (comment), you could have a workaround. Maybe a bit too manual, but better than nothing.

I think `ret.update` may not be enough. For my use-case, I would need to also remove some keys (`type`) of `ret`. I guess I'll have to resort to monkey patching for now.

PR #20, opened 2018-11-04
Fix response decorator return value, return a flask.Response object r…

aaronhill:
…ather than a tuple of (Response, <code>)


aaronhill:
> Thanks. I think I'll do it.

Sounds good, I'll send the PR your way and let you decide on merging.

lafrech:
Thanks. I had done the change in https://github.com/Nobatek/flask-rest-api/commit/25032b797c16ef4bd403c078560d6819194359c2 in a development branch so I picked it from there to avoid a merge conflict.

PR #22, opened 2018-11-07
Allow view create Response object

xen:
Allow view to create custom Response object.

### DIFF
--- a/docs/response.rst
+++ b/docs/response.rst
@@ -57,3 +57,31 @@ If a view function returns a list of objects, the :class:`Schema
    ``200`` code, decorating it with 
    :meth:`Blueprint.response <Blueprint.response>` is useful anyway, to return
    a proper Flask :class:`Response <flask.Response>` object.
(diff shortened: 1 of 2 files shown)


xen:
Another approach can be support for second value in return with code value. Like this:

```
@blp.response(code=204):
def delete(self, pet_id):
    pet = Pet.get_by_id(pet_id)
    if pet is None:
        return None, 404
    Pet.delete(pet_id)
```

But one that I've implemented gives more flexibility.

lafrech:
Thanks. I intend to do something like this.

I'd like to address https://github.com/Nobatek/flask-rest-api/issues/18 at the same time.

Basically, allow `Blueprint.response` to receive either a response and headers, or just a response, and the response could be just a payload or a `Response` object.

I just need to get my head around this.

lafrech:
@xen, I just sent another PR implementing this (#40). Feedback welcome.

PR #24, opened 2018-11-16
Prevent registering several times the same endpoint

lafrech:
Registering several routes / view functions for the same endpoint leads to all kinds of trouble.

The user may specify an optional endpoint name when calling `Blueprint.route` so he can avoid this. But if he does not, a unique endpoint name is provided.

This change allows to remove a few awkward statements that were meant to account for the multiple rules per endpoint case.

### DIFF
--- a/flask_rest_api/blueprint.py
+++ b/flask_rest_api/blueprint.py
@@ -27,7 +27,6 @@ Documentation process works in several steps:
 """
 
 from collections import OrderedDict
-from copy import deepcopy
 
 from flask import Blueprint as FlaskBlueprint
 from flask.views import MethodViewType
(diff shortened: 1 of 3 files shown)


lafrech:
This PR introduces a test that calls `route` twice on the same function with different decorators applied.

```py
        blp.route('/route_1')(blp.doc(description='Route 1')(func))
        blp.route('/route_2')(blp.doc(description='Route 2')(func))
```

As long as the decorators store doc information in the view func objects, this is wrong, because all decorators (here the two `blp.doc`) write to the same `dict`, so they interfere with each other.

This test is modified in master. The change in the PR is still correct.

It would be nice to find a better way to pass information than writing in the function objects, but an alternative is still to be found. Meanwhile, it is not such a big deal as the test above is a corner case that probably doesn't happen in real life.

PR #28, opened 2018-11-21
Allow multiple flask instances per Api instance

### LINKED ISSUE #27: Cannot create a second flask application
Hi,

I believe that due to recent changes around the Blueprint.doc decorator it's not possible to have 2 flask applications created at the same time. See below a minimal example. Tests should pass with flask-rest-api 0.10 but the test `test_app` fails with 0.11. Is this the expected behavior or it's a bug?

```
### app.py ###

from flask import Flask
from flask.views import MethodView

from flask_rest_api import Api, Blueprint

rest_api = Api()

_ = Blueprint('Test', __name__, url_prefix='/api/test')

@_.route('/')
class TestView(MethodView):

    @_.response()
    def get(self):
        return [1,2,3]

def create_app():
    app = Flask(__name__)

    rest_api.init_app(app)
    rest_api.register_blueprint(_)

    return app
```

```
### test_app.py ###

from app import create_app

def test_app():
    app = create_app()
    
    # This will fail with flask-rest-api >= 0.11
    app2 = create_app()
```

Here's the output from `pytest`
```
$ pytest                                                                                           
============================================================================ test session starts ============================================================================
platform darwin -- Python 3.7.0, pytest-4.0.0, py-1.7.0, pluggy-0.8.0
rootdir: /Users/sancho/Development/test, inifile:
collected 1 item

test_app.py F                                                                                                                                                         [100%]

================================================================================= FAILURES ==================================================================================
_________________________________________________________________________________ test_app __________________________________________________________________________________

    def test_app():
        app = create_app()
>       app2 = create_app()

test_app.py:5:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
app.py:21: in create_app
    rest_api.register_blueprint(_)
../flask-rest-api/flask_rest_api/__init__.py:81: in register_blueprint
    blp.register_views_in_doc(self._app, self.spec)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <flask_rest_api.blueprint.Blueprint object at 0x103a3d470>, app = <Flask 'app'>, spec = <flask_rest_api.spec.APISpec object at 0x10433a400>

    def register_views_in_doc(self, app, spec):
        """Register views information in documentation

        If a schema in a parameter or a response appears in the spec
        `definitions` section, it is replaced by a reference to its definition
        in the parameter or response documentation:

        "schema":{"$ref": "#/definitions/MySchema"}
        """
        # This method uses the documentation information associated with the
        # endpoint (in self._docs) to provide documentation for the route to
        # the spec object.
        for endpoint, doc in self._docs.items():
            # doc is a dict of documentation per method for the endpoint
            # {'get': documentation, 'post': documentation,...}

            # Prepend Blueprint name to endpoint
            endpoint = '.'.join((self.name, endpoint))

            # Format operations documentation in OpenAPI structure
            # Tag all operations with Blueprint name
            # Merge manual doc
            for key, (auto_doc, manual_doc) in doc.items():
                self._prepare_doc(auto_doc, spec.openapi_version)
>               auto_doc['tags'] = [self.name]
E               TypeError: 'str' object does not support item assignment

../flask-rest-api/flask_rest_api/blueprint.py:161: TypeError
```

lafrech:
Closes #27.

TODO:

- [ ] Changelog
- [ ] Documentation

### DIFF
--- a/flask_rest_api/spec/__init__.py
+++ b/flask_rest_api/spec/__init__.py
@@ -3,7 +3,6 @@
 import json
 
 import flask
-from flask import current_app
 import apispec
 
 from .plugins import FlaskPlugin
@@ -72,14 +71,14 @@ def _add_leading_slash(string):
 class DocBlueprintMixin:
     """Extend Api to serve the spec in a dedicated blueprint."""
 
-    def _register_doc_blueprint(self):
+    def _register_doc_blueprint(self, app):
         """Register a blueprint in the application to expose the spec
 
         Doc Blueprint contains routes to
         - json spec file
         - spec UI (ReDoc, Swagger UI).
         """
-        api_url = self._app.config.get('OPENAPI_URL_PREFIX', None)
+        api_url = app.config.get('OPENAPI_URL_PREFIX', None)
         if api_url is not None:
             blueprint = flask.Blueprint(
                 'api-docs',
@@ -87,18 +86,30 @@ class DocBlueprintMixin:
                 url_prefix=_add_leading_slash(api_url),
                 template_folder='./templates',
             )
-            # Serve json spec at 'url_prefix/openapi.json' by default
-            json_path = self._app.config.get(
-                'OPENAPI_JSON_PATH', 'openapi.json')
-            blueprint.add_url_rule(
-                _add_leading_slash(json_path),
-                endpoint='openapi_json',
-                view_func=self._openapi_json)
-            self._register_redoc_rule(blueprint)
-            self._register_swagger_ui_rule(blueprint)
-            self._app.register_blueprint(blueprint)
+            self._register_openapi_json_rule(app, blueprint)
+            self._register_redoc_rule(app, blueprint)
+            self._register_swagger_ui_rule(app, blueprint)
+            app.register_blueprint(blueprint)
 
-    def _register_redoc_rule(self, blueprint):
+    def _register_openapi_json_rule(self, app, blueprint):
+        """Serve json spec file"""
+        json_path = app.config.get('OPENAPI_JSON_PATH', 'openapi.json')
+
+        def openapi_json():
+            """Serve JSON spec file"""
+            # We don't use Flask.jsonify here as it would sort the keys
+            # alphabetically while we want to preserve the order.
+            return app.response_class(
+                json.dumps(self._specs[app].to_dict(), indent=2),
+                mimetype='application/json')
+
+        blueprint.add_url_rule(
+            _add_leading_slash(json_path),
+            endpoint='openapi_json',
+            view_func=openapi_json)
+
+    @staticmethod
+    def _register_redoc_rule(app, blueprint):
         """Register ReDoc rule
 
         The ReDoc script URL can be specified as OPENAPI_REDOC_URL.
@@ -115,12 +126,12 @@ class DocBlueprintMixin:
 
         OPENAPI_REDOC_VERSION is ignored when OPENAPI_REDOC_URL is passed.
         """
-        redoc_path = self._app.config.get('OPENAPI_REDOC_PATH')
+        redoc_path = app.config.get('OPENAPI_REDOC_PATH')
         if redoc_path is not None:
-            redoc_url = self._app.config.get('OPENAPI_REDOC_URL')
+            redoc_url = app.config.get('OPENAPI_REDOC_URL')
             if redoc_url is None:
                 # TODO: default to 'next' when ReDoc 2.0.0 is released.
-                redoc_version = self._app.config.get(
+                redoc_version = app.config.get(
                     'OPENAPI_REDOC_VERSION', 'latest')
                 # latest or v1.x -> Redoc GitHub CDN
                 if redoc_version == 'latest' or redoc_version.startswith('v1'):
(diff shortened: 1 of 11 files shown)


flask_rest_api/spec/__init__.py
```diff
@@ -132,13 +140,18 @@ def _register_redoc_rule(self, blueprint):
                     redoc_url = (
                         'https://cdn.jsdelivr.net/npm/redoc@'
                         '{}/bundles/redoc.standalone.js'.format(redoc_version))
-            self._redoc_url = redoc_url
+
```
svidela, line 146: I'm not sure why you moved this function here. I think you could use `current_app.name` as in `_openapi_json`. Same thing goes for `openapi_swagger_ui` below.

flask_rest_api/spec/__init__.py
```diff
@@ -87,24 +87,29 @@ def _register_doc_blueprint(self, app):
...
-        """Serve JSON spec file"""
-        # We don't use Flask.jsonify here as it would sort the keys
-        # alphabetically while we want to preserve the order.
-        return current_app.response_class(
-            json.dumps(self.spec.to_dict(), indent=2),
-            mimetype='application/json')
+    def _register_openapi_json_rule(self, app, blueprint):
+        """Serve json spec file"""
+        json_path = app.config.get('OPENAPI_JSON_PATH', 'openapi.json')
+
```
svidela, line 97: Similar to my other comment, I think you could use `current_app` instead of a closure with `app`. Except that here you may need to use: `current_app._get_current_object()` to access `self.specs`

flask_rest_api/spec/__init__.py
lafrech: Thanks for reviewing.

I also need to pass `redoc_url`, `swagger_ui_url` and `swagger_ui_supported_submit_methods`.

I find it easier to use a closure than to pass them via the app object.

Or maybe I misunderstood your suggestion.

Are you suggesting I leave them as private attributes to the spec object (which always looked patchy to me, BTW) and I pass the spec object via the app object as `app.extensions['flask-rest-api']['spec']` ?

flask_rest_api/spec/__init__.py
lafrech: Yes, perhaps I could store the spec object in `app.extensions['flask-rest-api']['spec']` and access it from there.

I don't have a strong feeling about it. Any downside with the closure approach?

flask_rest_api/spec/__init__.py
svidela: Sure. I haven't noticed those, sorry. I don't have a strong argument towards one approach or the other (I'm still familiarizing with your code).

flask_rest_api/spec/__init__.py
svidela: I don't have a strong argument against the closure approach. I just find a bit confusing having references to both `current_app` and `app` in the same function. I would pick one way to reference the application instance. Or maybe I'm missing something (?)

lafrech: Yeah, I agree it looks strange to use both. As I mentioned in the comment, I was using `current_app` here on the wrong assumption that `current_app.response_class` might be resolved at function definition time and would be wrong if the response class is changed after `Api.init_app`. I wan't sure about that but was too sloppy to check.

Here's the check:

```py
def outside(obj):
    def inside():
        return obj.attr
    return inside

class Obj:
    def __init__(self, val):
        self.attr = val

o = Obj(12)
f = outside(o)
assert f() == 12
o.attr = 42
assert f() == 42
```

Great. So I can change that to 

```py
            return app.response_class(
                json.dumps(self.specs[app].to_dict(), indent=2),
                mimetype='application/json')
```

svidela: Nice!

lafrech:
This POC looks functional, but I'm still worried it might be opening the door to corner case issues. It changes the philosophy of the library.

I'd like to see a convincing use case before pushing this further.

Holding this for now.

svidela:
Awesome! Thanks for considering and coding this. 

Regarding use cases (I'm thinking at loud here...), I think it would be nice to have this in case you want to implement some kind of application dispatching combining various Flask application instances (http://flask.pocoo.org/docs/1.0/patterns/appdispatch/#application-dispatching). I will try to code some examples to see how that would actually work.

cheers!

lafrech:
Closing this for now.

See https://github.com/Nobatek/flask-rest-api/issues/38#issuecomment-495527421.

PR #32, opened 2018-11-27
Allow adding definitions before registering app

lafrech:
Based on https://github.com/Nobatek/flask-rest-api/pull/28.

With current code, `definition` decorator only works if an app is already registered. This means the schema (and therefore the views) must be imported at app init time. Worse, if multiple apps are used, it will only work if definitions are added after all apps are initiated.

This change allows definitions to be registered before apps are initialized, i.e. at application import time, rather than init time. It allows a single file app with a `create_app` function.

 It keeps the old behaviour, so definitions can still be added after apps are initialized. This is mandatory if the `init_app` pattern is not used (if calling `Api(app)`), for instance.

TODO:

- [ ] #28 
- [ ] Changelog
- [ ] Documentation ?

### DIFF
--- a/flask_rest_api/__init__.py
+++ b/flask_rest_api/__init__.py
@@ -28,6 +28,8 @@ class Api(DocBlueprintMixin, ErrorHandlerMixin):
     """
     def __init__(self, app=None, *, spec_kwargs=None):
         self._specs = {}
+        # Use list rather than dict to enforce definition order
+        self._definitions = []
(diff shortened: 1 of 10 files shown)


lafrech:
This feature is achieved in https://github.com/Nobatek/flask-rest-api/commit/8d7967d1a2eb5e4cf39312d525687b807b755a10. We'll update if we ever finalize #28.

PR #54, opened 2019-03-31
Support `defaults` in `route`

zedrdave:
Small fix that allows the use of `defaults` in a route:
```
@ns.route('/<int:pet_id>/<int:version>', methods=['GET'])
@ns.route('/<int:pet_id>/', defaults={'version': None}, methods=['GET'])
class Pet(MethodView):
```

### DIFF
--- a/flask_rest_api/spec/plugins.py
+++ b/flask_rest_api/spec/plugins.py
@@ -59,6 +59,8 @@ class FlaskPlugin(BasePlugin):
         """Get parameters from flask Rule"""
         params = []
         for argument in rule.arguments:
+            if argument in rule.defaults:
+                continue
             param = {
                 'in': 'path',
                 'name': argument,


lafrech:
Sorry for the delay. I'm a bit reluctant about this. This wouldn't document the default value.

I don't think the "multiple routes per view function" is well supported in apispec/flask-rest-api so I wouldn't advocate for using it. (If I'm wrong about this, then great.)

If your use case is to have a GET method that returns an item if an ID is specified and a list otherwise, I'd suggest using two different view functions. I don't think it is such a bad limitation because when using a common function, it generally starts with a test (`if pet_id is None`) and then splits up into two branches with different return value types (and then I don't know how to document the fact that depending on whether an ID is provided, the return value may be singular or plural).

See example [here](https://github.com/marshmallow-code/apispec-webframeworks/issues/14#issuecomment-435592207) or in [@svidela's example](https://github.com/svidela/flask-rest-api-example/tree/master/app/resources) for an application using `MethodView` in a way that sticks close to the API.

What do you think? Would you like to elaborate about your use case?

zedrdave:
@lafrech The goal is definitely not to return different number of resources (which I agree, would be better served with separate endpoints).

The above example is the Flask-recommended way (that I know of) to have a route with a default value for `version`. Both `<pet_id>` and `<pet_id>/<version>` are accepted (with the former assigning a default value to `version`).

The change is really minimal and lets Flask handle it as it does with regular MethodView routes. As far as I can tell, it works as expected (with flask-rest-api).

What would be an alternative way to achieve this without supporting `defaults`?

lafrech:
> @lafrech The goal is definitely not to return different number of resources (which I agree, would be better served with separate endpoints).

Good. Sorry, just a random guess.

> The change is really minimal and lets Flask handle it as it does with regular MethodView routes. As far as I can tell, it works as expected (with flask-rest-api).

It works but it doesn't document the default and the parameter is marked as required.

I should also double-check that calling twice the `@route` on the same function doesn't do any wrong as this is not something that we explicitly support and test.

In fact, we can't document the parameter default because in OpenAPI, path parameters are always required. https://swagger.io/docs/specification/describing-parameters/#path-parameters

> What would be an alternative way to achieve this without supporting defaults?

Basically, in terms of REST API, those are two different routes. So perhaps a clean way to do that is to have two different view functions. You can still factorize the common part in a third method.

Perhaps flask-rest-api could help by documenting two separate resources. Or is it already what happens (due to the two calls to `@route`)? In this case, I totally missed that and your fix could be good, I guess. Can you please post the spec generated by your code?

Similar issue in flask-restplus: https://github.com/noirbizarre/flask-restplus/issues/381.

zedrdave:
I checked and indeed two separate resources are documented in Swagger. I can't post a clean example at the moment (my own API is probably a bit too complex to be useful here), but it should be easy to edit existing ones.

The only thing missing, would be a mention of the default in the one endpoint where it is not provided (probably not as trivial a change as this fix), but I would say having this feature supported would already be helpful, especially for compatibility with MethodView in general…

This could of course be achieved by calling a third function, but this would add a lot of verbosity (to what is already a fairly verbose API), especially in cases where many endpoints might have an optional parameter (such as a `version` one).

As for `flask-restplus`: it did support this exact code, which is why needed it in order to transition my code from it over to `flask-rest-api`.

lafrech:
Alright. Then I suppose this fix is on the right track.

From a quick glance, it looks like calling `@route` twice is not an issue thanks to the deepcopy at https://github.com/Nobatek/flask-rest-api/blob/master/flask_rest_api/blueprint.py#L162. I shall check this more thoroughly, add deep copies where needed and add comments to avoid removing those in the future. I remember I once removed code related to multiple routes per view because we didn't support this case as it lead to several issues, but this was a while ago. I need to review that.

Can you please add a test that checks the code passes and the generated doc is correct?

Thanks.

lafrech:
I fixed this in https://github.com/Nobatek/flask-rest-api/pull/58. Thanks for reporting the issue.

Calling `@route` twice is supported. In fact, it is explicitly tested ([`test_blueprint_multiple_routes_per_view`](https://github.com/Nobatek/flask-rest-api/blob/master/tests/test_blueprint.py#L455)).

> The only thing missing, would be a mention of the default in the one endpoint where it is not provided

I'm not sure I understand. I think there is no need to document the default. The fact that the view function expects a parameter and the route passes it as default is internal detail. From the API perspective, the two routes could be served by distinct (although quite similar) functions.

Overall, I see no shortcoming to this fix.

lafrech:
(What does not work right now is calling route twice on the same view function decorated with different decorators. This is because decorators store their doc in an attribute of the function object, so they conflict with each other.)

PR #72, opened 2019-05-10
Custom docstring ignore tag

### LINKED ISSUE #49: Customizable docstring ignore tag
# Summary

The string that stops docstring parsing (`---`) should be customizable.

# Details

According to [the docs](https://flask-rest-api.readthedocs.io/en/stable/openapi.html#add-documentation-information-to-the-view-functions), "the part of the docstring following the `'---`` line is ignored."

This is normally great, but can cause issues when using some docstring styles such as [NumPyDoc](https://numpydoc.readthedocs.io/en/latest/format.html#sections).

The separator value should be user-customizable via flask config value, perhaps `OPENAPI_DOCSTRING_SEP` or something like that.

## Current Behavior

```python
def get(...):
    """
    Find pets by ID.

    Return pets based on ID.

    Parameters
    ----------
    pet_id : int
        The PK for the pet to query

    Returns
    -------
    dict : :class:`orm.Pet`
        A dict representation of the pet instance. When returned via flask, is
        passed through :func:`flask.jsonify`.
    """
```

will be mangled to:

```python
{
    'get': {
        'summary': 'Find pets by ID.',
        'description': 'Return pets based on ID.\n\nParameters',
    }
}
```

Note the "Parameters" string in the description.

## Expected Behavior

```python
app = Flask('My API')
app.config['OPENAPI_VERSION'] = '3.0.2'
app.config['OPENAPI_DOCSTRING_SEP'] = "Parameters"
api = Api(app)
```

```python
def get(...):
    """
    Find pets by ID.

    Return pets based on ID.

    Parameters
    ----------
    pet_id : int
        The PK for the pet to query

    Returns
    -------
    dict : :class:`orm.Pet`
        A dict representation of the pet instance. When returned via flask, is
        passed through :func:`flask.jsonify`.
    """
```

```python
{
    'get': {
        'summary': 'Find pets by ID.',
        'description': 'Return pets based on ID.',
    }
}
```

# Implementation Notes

Any string should be acceptable as a separator. In addition, the `None` value should also be accepted. If the separator is `None`, then the entire docstring is included in the swagger details.

It looks like this docstring parsing is done in [utils.load_info_from_docstring](https://github.com/Nobatek/flask-rest-api/blob/master/flask_rest_api/utils.py#L35). I haven't investigated much further than that, but it looks like it might be pretty easy to implement.

dougthor42:
This allows the user to customize the docstring ignore separator.

The custom separator is set during blueprint initialization:

```python
blp = Blueprint('foo', __name__, docstring_sep="~~~")
@blp.route('/', methods=['GET'])
def view_func():
    """Summary
    
    Descr
    ~~~
    This line and more ignored.
    """
    return "foo"

## Results in:
# summary: "Summary"
# description: "Descr"
```

```python
blp = Blueprint('foo', __name__, docstring_sep=None)
@blp.route('/', methods=['GET'])
def view_func():
    """Summary
    
    Descr
    ~~~
    This line is included.
    """
    return "foo"

## Results in:
# summary: "Summary"
# description: "Descr\n~~~\nThis line is included."
```

Fixes #49.

### DIFF
--- a/flask_rest_api/blueprint.py
+++ b/flask_rest_api/blueprint.py
@@ -66,12 +66,11 @@ class Blueprint(
         "files": "multipart/form-data",
     }
 
-    DOCSTRING_SEP = "---"
+    DOCSTRING_INFO_DELIMITER = "---"
 
     def __init__(self, *args, **kwargs):
 
         self.description = kwargs.pop('description', '')
-        self.docstring_sep = kwargs.pop('docstring_sep', self.DOCSTRING_SEP)
 
         super().__init__(*args, **kwargs)
 
@@ -136,16 +135,18 @@ class Blueprint(
         endpoint_manual_doc = self._manual_docs.setdefault(
             endpoint, OrderedDict())
 
-        def store_method_docs(method, function, sep='---'):
+        def store_method_docs(method, function):
             """Add auto and manual doc to table for later registration"""
-            # Get summary/description from docstring
-            # and auto documentation from decorators
+            # Get auto documentation from decorators
+            # and summary/description from docstring
             # Get manual documentation from @doc decorator
-            docstring = function.__doc__
-            auto_doc = {}
-            if docstring:
-                auto_doc = load_info_from_docstring(docstring, sep=sep)
-            auto_doc.update(getattr(function, '_apidoc', {}))
+            auto_doc = getattr(function, '_apidoc', {})
+            auto_doc.update(
+                load_info_from_docstring(
+                    function.__doc__,
+                    delimiter=self.DOCSTRING_INFO_DELIMITER
+                )
+            )
             manual_doc = getattr(function, '_api_manual_doc', {})
             # Store function auto and manual docs for later registration
             method_l = method.lower()
@@ -157,12 +158,12 @@ class Blueprint(
             for method in self.HTTP_METHODS:
                 if method in obj.methods:
                     func = getattr(obj, method.lower())
-                    store_method_docs(method, func, sep=self.docstring_sep)
+                    store_method_docs(method, func)
         # Function
         else:
             methods = options.pop('methods', None) or ['GET']
             for method in methods:
-                store_method_docs(method, obj, sep=self.docstring_sep)
+                store_method_docs(method, obj)
 
         endpoint_auto_doc['parameters'] = parameters
 
--- a/flask_rest_api/blueprint.py
+++ b/flask_rest_api/blueprint.py
@@ -66,9 +66,12 @@ class Blueprint(
         "files": "multipart/form-data",
     }
 
+    DOCSTRING_SEP = "---"
+
     def __init__(self, *args, **kwargs):
 
         self.description = kwargs.pop('description', '')
+        self.docstring_sep = kwargs.pop('docstring_sep', self.DOCSTRING_SEP)
 
         super().__init__(*args, **kwargs)
 
@@ -133,13 +136,15 @@ class Blueprint(
         endpoint_manual_doc = self._manual_docs.setdefault(
             endpoint, OrderedDict())
 
-        def store_method_docs(method, function):
+        def store_method_docs(method, function, sep='---'):
             """Add auto and manual doc to table for later registration"""
             # Get summary/description from docstring
             # and auto documentation from decorators
             # Get manual documentation from @doc decorator
             docstring = function.__doc__
-            auto_doc = load_info_from_docstring(docstring) if docstring else {}
+            auto_doc = {}
+            if docstring:
+                auto_doc = load_info_from_docstring(docstring, sep=sep)
             auto_doc.update(getattr(function, '_apidoc', {}))
             manual_doc = getattr(function, '_api_manual_doc', {})
             # Store function auto and manual docs for later registration
@@ -152,12 +157,12 @@ class Blueprint(
             for method in self.HTTP_METHODS:
                 if method in obj.methods:
                     func = getattr(obj, method.lower())
-                    store_method_docs(method, func)
+                    store_method_docs(method, func, sep=self.docstring_sep)
         # Function
         else:
             methods = options.pop('methods', None) or ['GET']
             for method in methods:
-                store_method_docs(method, obj)
+                store_method_docs(method, obj, sep=self.docstring_sep)
 
         endpoint_auto_doc['parameters'] = parameters
 
--- a/flask_rest_api/utils.py
+++ b/flask_rest_api/utils.py
@@ -32,19 +32,22 @@ def get_appcontext():
     return ctx.flask_rest_api
 
 
-def load_info_from_docstring(docstring):
+def load_info_from_docstring(docstring, sep="---"):
     """Load summary and description from docstring"""
     split_lines = trim_docstring(docstring).split('\n')
 
-    # Info is separated from rest of docstring by a '---' line
-    for index, line in enumerate(split_lines):
-        if line.lstrip().startswith('---'):
-            cut_at = index
-            break
-    else:
-        cut_at = index + 1
+    split_info_lines = split_lines
 
-    split_info_lines = split_lines[:cut_at]
+    if sep is not None:
+        # Info is separated from rest of docstring by a `sep` line
+        for index, line in enumerate(split_lines):
+            if line.lstrip().startswith(sep):
+                cut_at = index
+                break
+        else:
+            cut_at = index + 1
+
+        split_info_lines = split_lines[:cut_at]
 
     # Description is separated from summary by an empty line
     for index, line in enumerate(split_info_lines):
--- a/flask_rest_api/utils.py
+++ b/flask_rest_api/utils.py
@@ -32,31 +32,33 @@ def get_appcontext():
     return ctx.flask_rest_api
 
 
-def load_info_from_docstring(docstring, sep="---"):
-    """Load summary and description from docstring"""
+def load_info_from_docstring(docstring, *, delimiter="---"):
+    """Load summary and description from docstring
+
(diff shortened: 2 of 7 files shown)


flask_rest_api/utils.py
```diff
@@ -32,19 +32,23 @@ def get_appcontext():
...
     split_lines = trim_docstring(docstring).split('\n')
 
-    # Info is separated from rest of docstring by a '---' line
-    for index, line in enumerate(split_lines):
-        if line.lstrip().startswith('---'):
-            cut_at = index
-            break
-    else:
-        cut_at = index + 1
+    # Default case: sep is None, thus no 'info' lines to split out.
```
lafrech: None is not the default case.

lafrech:
Thanks for the PR.

Code looks good.

Regarding the test, I'd prefer a minimalist test that tests each case (explicit `None`, default `"---"` and custom string) on a dummy ad-hoc docstring. Real-life examples are nice but I don't think it is worth 100 lines of code in this case.

flask_rest_api/blueprint.py
```diff
@@ -61,6 +61,7 @@ class Blueprint(
     def __init__(self, *args, **kwargs):
 
         self.description = kwargs.pop('description', '')
+        self.docstring_sep = kwargs.pop('docstring_sep', '---')
```
dougthor42: + Is this where you were thinking of adding the arg?
+ Where are the docs for the Blueprint so that I can add it there?

flask_rest_api/blueprint.py
```diff
@@ -117,13 +118,15 @@ def _store_endpoint_docs(self, endpoint, obj, **kwargs):
...
+        def store_method_docs(method, function, sep='---'):
             """Add auto and manual doc to table for later registration"""
             # Get summary/description from docstring
             # and auto documentation from decorators
             # Get manual documentation from @doc decorator
             docstring = function.__doc__
-            auto_doc = load_info_from_docstring(docstring) if docstring else {}
+            auto_doc = {}
+            if docstring:
+                auto_doc = load_info_from_docstring(docstring, sep=sep)
```
dougthor42: I broke this out to multiple lines because it was failing the lint test

flask_rest_api/utils.py
dougthor42: Yeah, it's poorly worded. I actually mean something more like "assume that a separator is not found". I'll update it.

flask_rest_api/utils.py
```diff
@@ -32,19 +32,23 @@ def get_appcontext():
...
+    split_info_lines = split_lines
+
+    if sep is not None:
+        # Info is separated from rest of docstring by a '---' line
```
dougthor42: Comment needs updating to reflect custom sep

tests/test_blueprint.py
```diff
@@ -430,6 +430,60 @@ def view_func():
...
+        api.register_blueprint(blp)
+        spec = api.spec.to_dict()
+        path = spec['paths']['/test/']
+        for method in ('put', 'patch', ):
+            assert path[method]['summary'] == 'Summary'
+            assert path[method]['description'] == 'Long description'
+
+        response = client.put('/test/')
+        assert response.status_code == 200
+        assert response.json == {'Value': 'OK'}
```
dougthor42: I probably should remove all the extra asserts from these tests and only check the strings, yeah? I guess I kept them in there to make sure that adding the arg didn't mess anything else up.

Lots of code duplication though - I'm not happy about that.

tests/test_blueprint.py
```diff
@@ -430,6 +430,60 @@ def view_func():
...
+            Not Ignored
+            """
+            return jsonify({'Value': 'OK'})
+
+        api.register_blueprint(blp)
+        spec = api.spec.to_dict()
+        path = spec['paths']['/test/']
+        for method in ('put', 'patch', ):
+            assert path[method]['summary'] == 'Summary'
+            assert path[method]['description'] == 'Long description\n~~~\nNot Ignored'
```
dougthor42: line fails linting.

tests/test_utils.py
```diff
@@ -55,3 +55,103 @@ def test_load_info_from_docstring(self):
...
+            '     "units": {string, optional},\n'
+            '     "upper_limit": {float, optional},\n'
+            '     "lower_limit": {float, optional}\n'
+            '   }\n\n'
+            'Returns\n'
+            '-------\n'
+            '200 :\n'
+            '    Success. Returned JSON data has two keys: ``old_value`` and\n'
+            '    ``new_value``, each containing only the values of the\n'
+            '    :class:`orm.Metric` object *that were requested to be changed*.\n'
```
dougthor42: line fails linting

dougthor42:
> Regarding the test, I'd prefer a minimalist test that tests each case (explicit None, default "---" and custom string) on a dummy ad-hoc docstring. Real-life examples are nice but I don't think it is worth 100 lines of code in this case.

OK, I'll change that. Do you want to keep everything in that single test case with multiple asserts? Or would you rather a new test function for each? Eg:

```python
def test_load_from_docstring_default_sep():
def test_load_from_docstring_custom_sep():
def test_load_from_docstring_none_sep():
```

----

Other questions:
1. Do you have any preferences on squashing? If not, I'll rebase/squash this into 2 commits:
    1. the change to `load_info_from_docstring` and related tests
    2. the change to the `Blueprint` class and related tests
2. Do you want the changelog update as part of this MR? If so, what would you like? (I don't see an "unreleased" section or anything like that.) Or will you handle the changelog update?
3. Where else do docs need to be updated?

flask_rest_api/blueprint.py
lafrech: I was thinking of making it a class attribute the user could override in a subclass, like `HTTP_METHODS`.

```py
class Blueprint:

    DOCSTRING_SEP = '---'
```
I'd document this in docs/openapi.rst.

flask_rest_api/blueprint.py
```diff
@@ -136,12 +139,12 @@ def store_method_docs(method, function):
             for method in self.HTTP_METHODS:
                 if method in obj.methods:
                     func = getattr(obj, method.lower())
-                    store_method_docs(method, func)
+                    store_method_docs(method, func, sep=self.docstring_sep)
```
lafrech: Can't you access `self.docstring_sep` in `store_method_docs` itself?

flask_rest_api/utils.py
```diff
@@ -32,19 +32,23 @@ def get_appcontext():
...
     split_lines = trim_docstring(docstring).split('\n')
 
-    # Info is separated from rest of docstring by a '---' line
-    for index, line in enumerate(split_lines):
-        if line.lstrip().startswith('---'):
-            cut_at = index
-            break
-    else:
-        cut_at = index + 1
+    # Assume the separator is not found: simply prevents an 'else' block.
```
lafrech: Nitpick: Unnecessary comment.

lafrech:
> OK, I'll change that. Do you want to keep everything in that single test case with multiple asserts? Or would you rather a new test function for each?

I don't really mind. But thanks for asking.

> Do you have any preferences on squashing? If not, I'll rebase/squash this into 2 commits:
> - the change to load_info_from_docstring and related tests
> - the change to the Blueprint class and related tests

That's perfect.

> Do you want the changelog update as part of this MR? If so, what would you like? (I don't see an "unreleased" section or anything like that.) Or will you handle the changelog update?

Don't bother. I can do it.

I generally don't create an "unreleased" section. I just add the next version when needed.

> Where else do docs need to be updated?

I think updating relevant section in docs/openapi.rst (https://github.com/Nobatek/flask-rest-api/blob/master/docs/openapi.rst#add-summary-and-description) is enough.

Thanks. Sorry for the delay...

flask_rest_api/blueprint.py
dougthor42: No, at least not with the base I'm working from. `store_method_docs` is defined within `Blueprint._store_endpoint_docs` and does not accept the instance as an arg.

My current base is a1c2dfc, v0.15.0.

flask_rest_api/blueprint.py
dougthor42: Do you still want me to move it to a class attribute rather than (or in addition to) an initialization arg?

flask_rest_api/blueprint.py
lafrech: I'm not familiar with closures. I thought `self` would be accessible since it is defined in the scope of `Blueprint._store_endpoint_docs` where `store_method_docs` is defined.

Currently out of office, I'll try to check this and the rest when I come back.

flask_rest_api/blueprint.py
lafrech: Class attribute only would be consistent with the rest of the lib.

The separator will most probably be something the user will use consistently in all his `Blueprint`s so subclassing and using a class attribute seems nice to me.

`description` is obviously different because it differs in all instances.

dougthor42: Sounds good. I added it as a class attribute.

I also left the initialization arg so that users can do any of these:

```python
# No need to subclass
blp0 = Blueprint('test', __name__, url_prefix='test', docstring_sep="~~~")

# But you can if you want to
class MyBlueprint(Blueprint):
    DOCSTRING_SEP = "~~~"

blp1 = MyBlueprint('test', __name__, url_prefix='test')

# and you can still override your subclass on a per-instance bases if desired.
blp2 = MyBlueprint('test', __name__, url_prefix='test', docstring_sep="^^^")
```

If that's not to your liking please let me know and I'll remove the init arg completely.

lafrech:
Rebased, squashed (perhaps a little too much) and reworked. Let's merge.

Issue #46, opened 2019-03-19
Multipart requests


asyncee:
Hello!

I did not found standard way to describe a multipart request. Is it supported?

Currently i had to write something like this to let uploaded files work:

```
request_body_description = {
    "content": {
        "multipart/form-data": {
            "schema": {
                "type": "object",
                "properties": {"logo": {"type": "string", "format": "binary"}},
            }
        }
    }
}


@bp.route("/update-logo")
class UpdateLogo(MethodView):
    @bp.response(code=204)
    @bp.doc(requestBody=request_body_description)
    def post(self):
        file = flask.request.files["logo"]
        filename = secure_filename(file.filename)
        binary = file.read()
        
        ... do something with file ...
```

This code allows swagger to render input type="file" with name "logo" and send multipart request when executing request from web interface.

Am i missing something?

Thanks!

toonalbers:
I think this is a bug/unimplemented feature, although I have not had time to investigate it or report it properly.

Form data in OAS3 should be in the `requestBody['content']` description as you have indicated, however [form and file locations](https://github.com/marshmallow-code/apispec/blob/237c700145b563cafa4cc8c18f0e4bdf87d352ff/src/apispec/ext/marshmallow/openapi.py#L65) are currently mapped to a `formData` OpenApi location (which was correct for OAS2, but is not valid anymore).

I suspect it could be fixed [here](https://github.com/Nobatek/flask-rest-api/blob/f4c2dfe1b86eced7d7b17a25af96e8cd559a9ae1/flask_rest_api/blueprint.py#L196) to also move `formData` locations to the `requestBody`. I'm not sure how that code would distinguish between `multipart/form-data` and `application/x-www-form-urlencoded` though.

lafrech:
Thanks for the report and investigation, and sorry for the delay.

https://swagger.io/docs/specification/describing-request-body/
https://swagger.io/docs/specification/describing-request-body/file-upload/

AFAIU, apispec's mapping is correct for OAS2. When using OAS3, files and form should be sent as `requestBody`, not parameters, so the mapping is not used (so files/form mapping is not really wrong in the OAS3 case, it is just useless).

[...]

PR #89, opened 2019-08-13
allow for swagger oauth redirect

vjm:
alllow for swagger to redirect properly for oauth


vjm:
@lafrech let me know if you have questions about this -- would love to see this get merged!

lafrech:
Hi. Thanks for your contribution.

Could you please elaborate on the use case? (Let's assume I know very little about OAuth.)

Is the template something you crafted from scratch, or is it borrowed from swagger-ui or elsewhere? I'd rather avoid duplication, so it would be nice if we could just point to an external JS file like we do with swagger-ui already.

vjm:
The use case is to use Swagger's built-in oauth features which are currently not possible with the implementation built into this library. Here's some more details on how you can use OAuth in Swagger: https://swagger.io/docs/specification/authentication/oauth2/

The template is borrowed from here: https://github.com/swagger-api/swagger-ui/blob/master/dist/oauth2-redirect.html
I am not sure that it's easily possible to extract this and use it as an external javascript file -- I am open to suggestions though.

Looks like the build failed due to linting, I'll take a look at that.

lafrech:
Looks like there is no easy way not to copy the whole html file.

- https://github.com/nelmio/NelmioApiDocBundle/issues/1097
- https://stackoverflow.com/questions/49518868/oauth2-authorization-in-nelmioapidocbundle/49519134#49519134

I'd rather not include a copy of a part of swagger-ui in the code (to minimize maintenance issues, having to update it when it changes).

Until we find a way, you could do that in you app by subclassing `DocBlueprintMixin`, I suppose.

The only missing piece would be the added line in the swagger-ui template:

```js
    oauth2RedirectUrl: "{{ url_for('api-docs.openapi_swagger_ui_redirect',_external=True) }}",
```

Maybe we could add an application parameter option for that. In fact, in any case we would, to avoid hardcoding the path.

This, plus a section in the docs explaining how to proceed.

I'm not really familiar with swagger-ui (we use ReDoc). The [CDN](https://cdnjs.com/libraries/swagger-ui) says 

> Swagger UI is a dependency-free collection of HTML, JavaScript, and CSS assets that dynamically generate beautiful documentation from a Swagger-compliant API

but there is no HTML in the package.

Maybe we could include the html file. After all, we already have swagger_ui.html. Except this one looks really minimal.

I'm also surprised the whole html file is a JS script. Why didn't they use an external JS script file and import it in the HTML?

vjm:
I've updated the PR to remotely pull the template that matches the swagger UI version in order to not have to store the template.

I've also added some documentation -- please take a look and let me know if we can merge this in! Would love to get the new release of this.

vjm:
@lafrech let me know what you think and if we can merge this in and release!

lafrech:
Don't hold your breath.

I'd like to finish ongoing work on #46/#83 and release 0.17 first.

I don't think pulling the file from an external GH repo is a nice option either.

Perhaps adding the file to the repo is the way to go. I'd like to avoid having to update it regularly, or to manage compatibility issues with different swagger-ui versions. We should at least check how regularly it is modified upstream and how they deal with backward compatibility.

The point is I want to limit the features of this lib to the minimum, to minimize maintenance load, so I try not to open doors to issues by adding features that can be easily achieved in client code. Especially features I don't use and I am not familiar with.

Maybe we could rework this to conditionally add the missing line in the swagger-ui template and add a note to the docs explaining how to achieve OAuth in client code.

vjm:
> I don't think pulling the file from an external GH repo is a nice option either.
> Perhaps adding the file to the repo is the way to go. I'd like to avoid having to update it regularly, or to manage compatibility issues with different swagger-ui versions. We should at least check how regularly it is modified upstream and how they deal with backward compatibility.

It hasn't been updated since Feb 13, 2018 so I think it's pretty stable.

> Maybe we could rework this to conditionally add the missing line in the swagger-ui template and add a not to the docs explaining how to achieve OAuth in client code.

@lafrech I have submitted more updates based on your feedback, please take another look and let me know.

vjm:
@lafrech any thoughts here?

vjm:
@lafrech any thoughts?

lafrech:
Looks like this could be achieved with no change to the code here by passing the swagger `oauth2RedirectUrl` option.

https://github.com/nelmio/NelmioApiDocBundle/issues/1097#issuecomment-1079558547

No time to check/confirm right now.

PR #95, opened 2019-09-15
Remove usage of strict option

sloria:
Assume marshmallow 3


lafrech:
Thanks. And congratulations for your first contribution!

I removed the `Meta` class from the quickstart in 12c9f90.

PR #123, opened 2020-01-31
Refactor Blueprint._prepare_doc

lafrech:
The decorators write doc information in each view function `_apispec` attribute. Since the docs may vary according to the OAS version, `Blueprint._prepare_doc` does some modifications at init time when the version is known.

This is far from perfect.

- [x] First, it is a simplification. The decorators write docs in a structure close to OAS then `_prepare_doc` just does a few fixes. It would be more generic if the decorators wrote docs in an independent structure and then the doc would be generated at once when all parameters are known. ~This PR does not address that yet.~

- [x] Each decorator (each Mixin) should be responsible for his documentation. This PR fixes this by splitting the preparation in separate methods.

- [x] The OAS version is not the only parameter that can be discovered at init time. There could also be application parameter affecting docs, such as whether or not a feature (authentication, ETag,...) is active. This PR fixes this by passing app and spec to the preparation methods.

Users willing to extend flask-smorest may create their own `_prepare_x_doc` methods and add them to the list of processors when subclassing `Blueprint`.

We should go a little bit further and decide that each decorator puts its own stuff in a namespaced substructure (e.g. `arguments` in `_apispec["arguments"]`) and the final doc structure is written at init time. For a start, the initial substructure can be the current final-doc-like structure. No need for more abstraction. But there may be cases in other decorators where it makes more sense to store parameters first then write doc than to first write doc then fix it. Without the namespacing, decorators all write temporary doc in the same structure so there could be interference between them.

### DIFF
--- a/flask_smorest/arguments.py
+++ b/flask_smorest/arguments.py
@@ -4,6 +4,7 @@ from functools import wraps
 
 from webargs.flaskparser import FlaskParser
 
+from .utils import deepupdate
 from .spec import DEFAULT_REQUEST_BODY_CONTENT_TYPE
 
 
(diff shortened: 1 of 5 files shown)


lafrech:
Shortcomings in OP are addressed. Should be good to go.

PR #124, opened 2020-01-31
Auto document errors, pagination, ETag

lafrech:
This PR brings automatic documentation of

- default errors
- pagination feature
- ETag feature (WIP)

It is based on #123 as the _prepare_doc rework will be needed to document ETag correctly.

I've been testing this on an app of ours and the result in ReDoc was neat. It still needs a bit of work on ETag to ensure it is documented only if enabled.

There are things I'm not 100% happy about but that's life. I've been holding this for month and it is better than nothing, yet.

One of my concerns is the mixins not being totally independent but there's nothing to do about that, I'm afraid.

Feedback welcome.

### DIFF
--- a/flask_smorest/__init__.py
+++ b/flask_smorest/__init__.py
@@ -1,5 +1,7 @@
 """Api extension initialization"""
 
+import http
+
 from webargs.flaskparser import abort  # noqa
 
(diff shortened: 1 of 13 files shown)


lafrech:
Superseded by #125.

PR #132, opened 2020-02-27
Move to webargs 6.0.0

### LINKED ISSUE #117: TypeError: use_args() got an unexpected keyword argument 'locations'
```
    def decorator(func):
    
        @wraps(func)
        def wrapper(*f_args, **f_kwargs):
            return func(*f_args, **f_kwargs)
    
        # Add parameter to parameters list in doc info in function object
        # The deepcopy avoids modifying the wrapped function doc
        wrapper._apidoc = deepcopy(getattr(wrapper, '_apidoc', {}))
        wrapper._apidoc.setdefault('parameters', []).append(parameters)
    
        # Call use_args (from webargs) to inject params in function
        return self.ARGUMENTS_PARSER.use_args(
>           schema, locations=[location], **kwargs)(wrapper)
E       TypeError: use_args() got an unexpected keyword argument 'locations'

../../../.virtualenvs/jAjXlZQO/lib/python3.7/site-packages/flask_smorest/arguments.py:76: TypeError
```

lafrech:
Closes #117.

TODO:

- [x] Modify pagination schema as unknown=IGNORE to avoid 422ing when other query args are passed on marshmallow 3.
- [x] Investigate consistency between namespaced webargs validation errors and errors triggered from view func.

### DIFF
--- a/docs/arguments.rst
+++ b/docs/arguments.rst
@@ -76,6 +76,64 @@ view function.
         def post(pet_data, query_args):
             return Pet.create(pet_data, **query_args)
 
+Multiple Arguments Schemas
+--------------------------
+
+To define arguments from multiple locations, calls to ``arguments`` decorator
+can be stacked:
+
+.. code-block:: python
+    :emphasize-lines: 4,5
+
+    @blp.route('/')
+    class Pets(MethodView):
+
+        @blp.arguments(PetSchema)
+        @blp.arguments(QueryArgsSchema, location="query")
+        @blp.response(PetSchema, code=201)
+        def post(self, pet_data, query_args):
+            pet = Pet.create(**pet_data)
+            # Use query args
+            ...
+            return pet
+
+It is possible to define multiple arguments for the same location. However,
+with marshmallow 3, schemas raise by default on unknown fields, so they should
+be tweaked to define ``unknown`` meta attribute as ``EXCLUDE``.
+
+.. code-block:: python
+    :emphasize-lines: 5,12
+
+    # With marshmallow 3, define unknown=EXCLUDE
+    class QueryArgsSchema1(ma.Schema):
+        class Meta:
+            ordered = True
+            unknown = ma.EXCLUDE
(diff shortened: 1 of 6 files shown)


lafrech:
With webargs 6, errors messages are namespaced by location. Since `use_args` decorators execute independently and each call will raise on error, the returned error message can't contain errors from several locations. In other words, if there are errors in both query args and body, and query args are parsed first, only errors from query args will be returned, then body errors once query args are fixed.

So first

    {"query": {"arg_1": [...], "arg_2": [...]}

then

    {"json": {"field_1": [...], "field_2": [...]}}

Errors raised from inside the view are not namespaced. The error message is what the user enters as `"errors"` parameter to `abort` (it can be a string, a dict, a list, whatever):

    abort(422, errors=...)

Of course, the user can sort of namespace manually by entering whatever he wants as error structure:

    abort(422, errors={"database": {"id": "Field must be unique"}})

Namespacing makes sense for webargs validation errors as errors are triggered from a specific location. In fact it is even necessary to discriminate two arguments with same name and different location. But there is no natural namespace for other errors.

Although it might look inconsistent, it is not an issue. This error structure won't be merged with an error structure from webargs validation since webargs validation errors prevent from entering the view.

Related webargs PRs:

https://github.com/marshmallow-code/webargs/pull/420
https://github.com/marshmallow-code/webargs/pull/463

mekanix:
Hello!

First of all, thank you for your swift response about upgrade to webargs 6.0. As FreeBSD maintainer of webargs (and flask-smorest) I updated webargs to 6.0 and now flask-smorest is broken. By no means do I try to rush the solution, but I have to ask if there's any ETA? If you expect to release new version of flask-smorest in next few days, I'll leave the ports/packages as they are, but if not, I would like to revert webargs to latest 5 version so flask-smorest is not broken for users.

lafrech:
Thanks for the packaging work.

I shall do this soonish.

PR #140, opened 2020-03-27
remove inline script (csp)

ev-agelos:
Move javascript code from html files into their own .js files.
This will allow servers that have set "unsafe-inline" in their content security policy header to run the js code and render the swagger docs.

My use case is what I described above, I have set the Content Security Policy header for scripts to "unsafe-inline" and it did not let any inline js code to be run.

I am not expert in html/js so please let me know if I need to change anything.
I also did not lowered the swagger submitted methods as I saw the return value was already lowered, please let me know if this is not always true.

### DIFF
--- a/flask_smorest/spec/__init__.py
+++ b/flask_smorest/spec/__init__.py
@@ -39,6 +39,7 @@ class DocBlueprintMixin:
                 __name__,
                 url_prefix=_add_leading_slash(api_url),
                 template_folder='./templates',
+                static_folder='static',
             )
             # Serve json spec at 'url_prefix/openapi.json' by default
             json_path = self._app.config.get(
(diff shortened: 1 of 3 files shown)


lafrech:
Sorry, I saw this, but didn't answer yet.

While ReDoc is easy to expose, swagger-ui is more of a pain (see open issues). I'd like to find a way to publish it with minimal maintenance burden.

I didn't find the time to dive into it yet.

There's a flask-swagger lib out there that packages swagger-ui. Maybe it addresses your issue.

Perhaps we should defer swagger-ui support to this lib (as an optional dependency).

ev-agelos:
Hey no worries, It may be a headache to support swagger-ui.
Thanks for the suggestion I will look into it.
I really like your plugin, it was exactly what I was looking for, so easy and nice to define args/schemas etc!
Feel free to reject the pr and Keep it up!

PR #154, opened 2020-05-31
Add "flask openapi print" command

lafrech:


lafrech:
This works fine unless the application prints other stuff to stdout (spurious prints during init, debug info,...).

Maybe we should add a file path argument.

lafrech:
I could cast the `pathlib.Path` to `str` to get the Python 3.5 tests to pass, but let's just drop Python 3.5 (#155).

PR #159, opened 2020-06-08
Add Blueprint.alt_response decorator to document alternative responses

### LINKED ISSUE #44: Error response schema
I've seen issue #19 and the recomendation instead of having more than one response is to use `@blp.doc`

The issue with this a approach is that it doesn't support any Schema information.

If we add schema as the object it doesn't work because it's not being replaced by it's json-schema's $ref
```py
@blp.doc(responses={
        '401': {'schema': GenericErrorSchema, 'description': 'Invalid credentials'}
    })
```

On the other hand when setting more than one `@blp.response` the openapispec is ok because it's using the autogenerated definition. The only issue is on the response serialization wrapper

Lot's of ideas come to my mind to fix this, but what do you think?
- Allow `@blp.response` with some extra param like `only_doc=True`
- Add `@blp.response_doc` to the `ResponseMixin` where `wrapper` just returns `result_raw`
- Allow using `@blp.doc(responses=...)` with GenericErrorSchema
- Make my `GenericErrorSchema` json encode into a `$ref` dict by calling something like [OpenAPIConverter.get_ref_dict](https://apispec.readthedocs.io/en/stable/api_ext.html#apispec.ext.marshmallow.openapi.OpenAPIConverter.get_ref_dict)

lafrech:
Closes #44.

Allows to document error with a schema like in response or with a reference.

Needs a bit of factorization.

Makes me want to change response signature to make code the first positional argument, for consistency. There is no real reason to use 200 as a default. (#193)

### DIFF
--- a/docs/response.rst
+++ b/docs/response.rst
@@ -52,3 +52,29 @@ If a view function returns a list of objects, the :class:`Schema
         @blp.response(200, PetSchema(many=True))
         def get(self, args):
             return Pet.get()
+
+.. note:: If a view function returns a :class:`werkzeug.BaseResponse`, that
+   response object is returned unchanged: it is not dumped by the schema and
+   the status code is not applied.
+
+.. note:: If a view function returns a tuple containing a status code, this
+   status code is used in place of the one specified as ``response`` parameter.
+   Doing this is generally a bad idea because the response status code won't
+   match the code in the API documentation.
+
+Documenting Alternative Responses
+=================================
+
+The :meth:`Blueprint.response <Blueprint.response>` decorator is meant to
+generate and document the response corresponding to the "normal" flow of the
+function. There can be alternative flows, if the function raises an exception,
+which results in a ``HTTPException``, or if it returns a ``Response`` object
+which is returned as is.
+
+Those alternative responses can be documented using the
+:meth:`Blueprint.response <Blueprint.alt_response>` decorator. Its signature is
+the same as ``response`` but its parameters are only used to document the
+response.
+
+A view function may only be decorated once with ``response`` but can be
+decorated multiple times with nested ``alt_response``.
--- a/flask_smorest/utils.py
+++ b/flask_smorest/utils.py
@@ -23,6 +23,11 @@ def deepupdate(original, update):
     return original
 
 
+def remove_none(mapping):
+    """Remove None values in a dict"""
+    return {k: v for k, v in mapping.items() if v is not None}
+
+
 # XXX: Does this belong here?
 def get_appcontext():
     """Return extension section from top of appcontext stack"""
--- a/tests/test_blueprint.py
+++ b/tests/test_blueprint.py
@@ -652,6 +652,172 @@ class TestBlueprint:
                 api.spec, 'response', error_response
             )
 
+    @pytest.mark.parametrize('openapi_version', ['2.0', '3.0.2'])
+    def test_blueprint_alt_response_schema(
+            self, app, openapi_version, schemas
+    ):
+        """Check alternate response schema is correctly documented"""
+        app.config['OPENAPI_VERSION'] = openapi_version
+        api = Api(app)
+        blp = Blueprint('test', 'test', url_prefix='/test')
+
+        example = {'error_id': 'E1', 'text': 'client error'}
+        examples = {
+            'example 1': {'error_id': 'E1', 'text': 'client error 1'},
+            'example 2': {'error_id': 'E2', 'text': 'client error 2'},
+        }
+        headers = {'X-Custom-Header': 'Header value'}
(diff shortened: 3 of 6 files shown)


tests/test_blueprint.py
```diff
@@ -652,6 +652,96 @@ def func():
...
+        def func():
+            pass
+
+        @blp.route('/description')
+        @blp.error(400, schemas.ClientErrorSchema, description='Client error')
+        def func_with_description():
+            pass
+
+        @blp.route('/example')
+        @blp.error(400, schemas.ClientErrorSchema, example=example)
```
ilx: can you add multiple @blp.errors?

lafrech: Yes absolutely. I ought to add a test to show it. Thanks for the feedback.

The decorator wraps the function, so if 10 error decorators are stacked, that means 10 nested function calls on each request. Not ideal but should be fast enough.

The wrapping is done to avoid modifying the function object itself which would yield wrong docs if people apply different decorators on the same view function. Not the main use case, but could happen.

If there is a perf penalty issue, we could create a decorator to document multiple errors at once. The syntax wouldn't be as nice but oh well. Let's not dive into unneeded optimization for now, anyway.

ilx: 👍

erdnaxeli:
Hi,

What is preventing this to progress?

lafrech:
See discussion in https://github.com/marshmallow-code/flask-smorest/pull/163#issuecomment-651388788.

lafrech:
Rebased on #193.

Feedback welcome.

tests/test_blueprint.py
lafrech: Tested and fixed. Thanks for the heads up.

erdnaxeli:
This looks good, but I am wondering about removing the default code value for the `response` decorator. I understand the goal to have a similar interface between the two decorators.

TitaniumHocker:
Are there any updates?

lafrech:
Well, not really. I was waiting for feedback on this and #193/#199. I didn't get much but at least no one complained so I shall merge this soonish.

My plan is to release #208 and #222 (rework of automatic response component registration) in 0.28, then this (alt_response) and #193 (code as positional parameter) in 0.29.

It's now just a matter of carving out time to ship each release.

erdnaxeli (approved):

flask_smorest/utils.py
```diff
@@ -23,6 +23,11 @@ def deepupdate(original, update):
     return original
 
 
+def remove_none(mapping):
+    """Remove None values in a dict"""
+    return {k: v for k, v in mapping.items() if v is not None}
```
erdnaxeli, line 28: You didn't test it.
```Python
def test_remove_none():
    mapping = {"a": 0, "b": None, "c": 3, "d": None, e: "", "f": 6, g: False}
    
    result = remove_none(mapping)

    assert result == {"a": 0, "c": 3", e: "", "f": 6, "g": False}
```

lafrech: No. Technically, the test coverage remains the same, it is just a refactor.

But sure, since this util function is now available to test, I could add an extra test.

docs/response.rst
```diff
@@ -52,3 +52,29 @@ If a view function returns a list of objects, the :class:`Schema
...
+=================================
+
+The :meth:`Blueprint.response <Blueprint.response>` decorator is meant to
+generate and document the response corresponding to the "normal" flow of the
+function. There can be alternative flows, if the function raises an exception,
+which results in a ``HTTPException``, or if it returns a ``Response`` object
+which is returned as is.
+
+Those alternative responses can be documented using the
+:meth:`Blueprint.response <Blueprint.alt_response>` decorator. Its signature is
```
neumantm, line 75: I think the first part should be `Blueprint.alt_response`. Maybe the second can be `Blueprint.response`.

zedrdave:
Currently, it seems that `alt_response` needs a schema for the alternative response. However, in many (most?) cases, these alt_response would be a `40x`, `30x` etc code, and often handled via a `bp.abort()` (with no need for a schema)…

What is the recommended way to handle this? use `@doc` instead?

lafrech:
`abort()` doesn't mean no content so a schema is still needed to document the response content, be it the default error response.

But I agree we should allow `None` for a "no content" response.

Creating issue #260 to keep track of this.

zedrdave:
Shouldn't the default error schema be part of the API then (or is it and I missed it)? Seems like an unnecessary repetition for all projects to redefine an error schema for HTTPException… no?

lafrech:
@zedrdave see ongoing `alt_response` rework in #283 and explanations in PR comment.

Issue #36, opened 2019-01-23
Add authorization to header?


dels07:
Currently I use flask_jwt_extended for API security, how can I define the need of passing Authorization header?

I tried using this:
`@blp.doc(parameters={'Authorization': {'name': 'Authorization', 'in': 'header', 'description': 'Authorization: Bearer <access_token>', 'required': 'true'}})`

but it throw an error 

> builtins.KeyError 
KeyError: 'in

lafrech:
I didn't try, but I think this syntax should avoid the error.

```py
    @blp.doc(parameters=[{'name': 'Authorization', 'in': 'header', 'description': 'Authorization: Bearer <access_token>', 'required': 'true'}])
```

However I think this is wrong, because this list of parameters will replace the list that is generated automatically rather than extend it.

I think the proper way to do that would be to create your own decorator, extending `@jwt_required`, that would append your parameter to `func._apidoc`. Check how the `arguments` does that (https://github.com/Nobatek/flask-rest-api/blob/master/flask_rest_api/arguments.py#L55).

It shouldn't be complicated. Tell me how it goes, and feel free to ask if this is unclear.

dels07:
Thanks for your suggestion @lafrech 

I'm still new to python, this is the code I tried

```py
def doc_jwt_required(arg):
    @wraps(func)
    def wrapper(func):
        @jwt_required
        def inner(*args, **kwargs):
            parameters = [{
                'name': 'Authorization',
                'in': 'header',
                'description': 'Authorization: Bearer <access_token>',
                'required': 'true'
            }]

            func._apidoc = getattr(func, '_apidoc', {})
            func._apidoc.setdefault('parameters', []).append(parameters)
            func(*args, **kwargs)

        return inner
    return wrapper
```

However I checked that now all my parameters are gone for the route that implementing above decorator, could you help me what's wrong with it

lafrech:
Maybe something like this would work (I didn't test):

```py
from functools import wraps

def doc_jwt_required():
    @wraps
    def wrapper(func):
        parameters = [{
            'name': 'Authorization',
            'in': 'header',
            'description': 'Authorization: Bearer <access_token>',
            'required': 'true'
        }]
        func._apidoc = getattr(func, '_apidoc', {})
        func._apidoc.setdefault('parameters', []).append(parameters)
        return jwt_required(func)
    return wrapper
```

Tell me if it works.

[...]

PR #163, opened 2020-06-20
Added error documenting to Blueprint.response

### LINKED ISSUE #44: Error response schema

RickLuiken:
Closes #44 

This is an alternative solution to #159 (of which I used a lot of code was used). This also fixes some of the positioning issues of [referenced here](https://github.com/marshmallow-code/flask-smorest/issues/36#issuecomment-461818791). I would love to hear whether people prefer this method or the method in #159.

Also, as multiple responses decorators can be used to document multiple errors, they could also be used to document multiple responses (with different codes) if needed. For this, it might be useful to check whether the status code specified by the response decorator matches the status code returned by the API method. I didn't include this as this might break some stuff.

### DIFF
--- a/flask_smorest/response.py
+++ b/flask_smorest/response.py
@@ -4,6 +4,7 @@ from copy import deepcopy
 from functools import wraps
 import http
 
+from werkzeug.exceptions import default_exceptions
 from werkzeug.wrappers import BaseResponse
 from flask import jsonify
 
@@ -19,13 +20,14 @@ class ResponseMixin:
     """Extend Blueprint to add response handling"""
 
     def response(
-            self, schema=None, *, code=200, description=None,
+            self, schema_or_ref=None, *, code=200, description=None,
             example=None, examples=None, headers=None
     ):
         """Decorator generating an endpoint response
 
-        :param schema: :class:`Schema <marshmallow.Schema>` class or instance.
-            If not None, will be used to serialize response data.
+        :param schema_or_ref: Either a :class:`Schema <marshmallow.Schema>`
+            class or instance or a string error reference.
+            When passing a reference, arguments below are ignored.
         :param int|str|HTTPStatus code: HTTP status code (default: 200).
             Used if none is returned from the view function.
         :param str description: Description of the response (default: None).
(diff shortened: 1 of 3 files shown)


lafrech:
Thank you @RickLuiken for your contribution.

I'd find it nice to have a single `response` decorator rather than add a new one as in #159. But I'm afraid there is no way to discriminate between

- documenting a possible answer
- documenting a possible answer and actually dumping the output with the schema

The case of passing a ref as string is obvious, but when passing a real schema, we can't tell whether we need to dump or not.

You're using werkzeug's `default_exceptions` because you make the assumption that the response is either a normal response with a 2xx code or an error with a 4xx or 5xx code. As I wrote in #166, calling it "error" is not accurate. The alternate response can be a 304, for instance, which is not an error. I could clarify this by renaming `error` to something like `alt_response` in #159.

There might be another way, however. I'm realizing that "errors" / "alternate responses", at least in my usage, are generated using `abort`, so the dump part of the wrapper is never reached. The response is generated in the error handler (which is also misnamed if the response is not really an error). Likewise, the dump would never occur either if the function returned a `Response` object rather than raising an `Exception`.

We could rely on this with two restrictions:

- The user shouldn't expect `response` to dump and generate the `Response` object for alternate responses. This case shall be covered by either the exception -> error handler path or the `Response` generated in the view func path. This sounds like poor design, but I don't think it is a blocker.
- The user would have to ensure that the "normal" response is the last decorator (the closest to the view function). I don't think this would be a blocker either, but is also sounds bad, hard to understand.

Another way to see it is to stop making a distinction between normal and alternate responses, and allow `response` to be applied several times for different responses (#166). I now understand this is what you meant by

> Also, as multiple responses decorators can be used to document multiple errors, they could also be used to document multiple responses (with different codes) if needed. For this, it might be useful to check whether the status code specified by the response decorator matches the status code returned by the API method.

That would be great. Except we don't expect the view function to return the status code. In fact, this is one of the goals of the `response` decorator, allowing the view function to just return content and letting the decorators do the rest (pagination, status code, ETag, etc...). Unless we change this and expect the view func to return the status code. This would be more generic, but it would also be a huge breaking change and it might make things complicated. I'm reluctant to do that for now.

We definitely have to chose between

- 1/ There can be multiple successful responses but the view func must return the status code.
- 2/ There can be only one successful response, the alternate responses use `abort` to shortcut `response` (or return a `Response` object). This is current behaviour.

 2/ allows to use `response` to document alternate responses, provided the successful `response` comes last in the decorator chain. This kinda breaks the zen. Might be better to have an explicit `alt_response` decorator.

I hope this is clear enough. Feedback welcome.

erdnaxeli:
Hi,

We are using Flask-Smorest at work and we bumped into this issue of documentating alternate responses (in our case, errors). We currently use `@bp.doc`, but it is not so great (we would like to use Marshamallow schemas) as you already noticed it. So here is my feedback.

First, I would have expected `@bp.response` to handle many response (with many decorators). I though it could differentiate them by status code as @RickLuiken suggest, but I also understand your point @lafrech. Maybe we could imagine a sort of hybrid behavior where `@bp.response` use the status code if available to differentiate responses, and if not it default to some response… but now the problem is to choose this default response. I think any solution we choose would have some corner cases not obvious and probably source of many errors, I don't really like it.

Looking at how [FastAPI does it](https://fastapi.tiangolo.com/advanced/additional-responses/), I think a 3rd option similar to what you are proposing with `alt_response` is better : 
* `@bp.alt_response` add the response to the doc, can generate the OpenAPI model from a Marshmallow Schema, like `response`
* `@bp.response` must be unique (triggering a warning could be great), like now
* `@bp.alt_response` can be repeated
* if we return a `Response`, Smorest does not do anything to it, like now
* else, Smorest tries to serialize it according to `@bp.response`, like now

So like in the FastAPI doc, the Smorest doc should make it clear that to return an alternate response you have to serialize yourself and return a `Response` object. That's not satisfying but as alternate responses are (I think) mostly errors, they will actually be formatted by a response handler and not in the view, so it's ok.

This avoid the "you must specify the successful response last in the decorator chain", which I find really confusing and I know in advance will cause many errors.

lafrech:
@erdnaxeli Thanks for the link to FastAPI. Indeed, I guess this is the most practical solution. It's not ideal (manual serialization in corner cases), but I agree it should be the best in most cases (errors).

We could add a new `alt_response` decorator (I'm not really satisfied with that name) or add a parameter to `response` to add a dict with needed info, like in FastAPI. The former is closer to what we actually do in flask-smorest and it makes the code look nicer, so I think we should go for that.

In fact, this is pretty close to what I did in #159. I shall  rename error into alt_response or something, and test multiple nested calls.

lafrech:
APIFairy also has a specific decorator for alternate responses: https://apifairy.readthedocs.io/en/latest/decorators.html#other-responses.

lafrech:
I reworked #159. Feedback welcome.

lafrech:
Let's move the discussion to #159.

Issue #164, opened 2020-06-22
Consider allow setting up OpenAPI title, not from app name


playpauseandstop:
Hi,

Thanks a lot for this library, I’ve switched to it recently from [flask-apispec](https://pypi.org/project/flask-apispec/) and really pleased of its api, proper OpenAPI 3 support, and overall developer experience. One more time, thanks a lot!

And I have one thing, that I’d like to discuss. Currently flask-smorest setting up `info.title` of OpenAPI 3 schema as Flask app name, but I’m not sure this is proper assumption. From my practice OAS 3 title is more likely human friendly string such as “Product API”, not an import name such as “product_api”.

With that in mind, would it be better to allow developer setup `API_TITLE` via app config as done for `API_VERSION`?

If you against that change, it still able to setup the title via `api.spec.title` assignment, but in that case I’d like to fix title context var for redoc or swagger ui templates, to reuse the value from the spec, instead of app name.

I would be glad to provide a PR for both suggestions if you’ll agree with proposed changes 

Thanks

lafrech:
Hi. Thanks for the nice feedback. I appreciate.

I do set a human title when instantiating the app:

```py
    app = Flask('Human friendly app name that appears in spec')
```

Example: https://github.com/sigopti/pyodhean-server/blob/master/pyodhean_server/app.py

It is hardcoded in my code, but I could use an envvar or a config param to change it. I don't have a strong objection to add this config option but I'm curious about the use case.

playpauseandstop:
Wow.

I must say I’m surprised with your code. Surprised in a good way. To be honest, I’ve not used Flask at all for last 5 or 6 years, so my knowledge about the topic is a bit limited at a moment, but I never know that first parameter to `Flask` instance can be a human friendly name. I always thought about it as import name. I’ve searched [docs](https://flask.palletsprojects.com/en/1.1.x/api/#application-object) and couldn’t find example similar to what you have used as docs rather explain necessity to have first argument as valid import name, so thanks a lot for sharing that snippet with me.

With that in mind, I even don’t know whether my initial request is valid or not now, but as you don’t have anything against I still would like to provide support for `API_TITLE` config var. I will be able to come up with PR in couple of next days.

Thanks

lafrech:
I'm afraid you're right and I've been misusing this parameter all along.

https://flask.palletsprojects.com/en/1.1.x/api/#application-object
[...]

PR #165, opened 2020-06-24
feat: Allow specify API title from config

playpauseandstop:
By default, API title still using Flask app name, but it is now possible to pass custom string via `API_TITLE` config var.

Also ensure using spec title within `<title>` tag at Swagger UI / Redoc pages.

Issue: #164


playpauseandstop:
Closing in favour of #169

Issue #65, opened 2019-05-03
Path parameters: document converters parameters


lafrech:
In FlaskPlugin, add min/max to number converters.

Manage negative values introduced in Werkzeug 0.15 (https://github.com/pallets/werkzeug/pull/1355).

lafrech:
Also see #170.

We need to replace the converter -> (type, format) mapping by a converter -> callable mapping.

PR #170, opened 2020-07-07
feat(Converters): Improve support of AnyConverter to list values

deytao:
Currently it is not possible to list values used in the `AnyConverter`.
It will help to have this functionality to ensure that the correct values are documented.

### DIFF
--- a/flask_smorest/spec/plugins.py
+++ b/flask_smorest/spec/plugins.py
@@ -65,11 +65,14 @@ class FlaskPlugin(BasePlugin):
                 'name': argument,
                 'required': True,
             }
-            type_, format_ = self.converter_mapping.get(
-                type(rule._converters[argument]), DEFAULT_TYPE)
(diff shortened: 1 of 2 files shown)


deytao:
Any thoughts?

lafrech:
Looks like we need to change the whole behaviour, here.

Current approach associates a converter with a (type, format) couple. We need to generalize this to use a callable, I guess, to generate the docs from the converter object.

This will allow to support other converters parameters:

https://werkzeug.palletsprojects.com/en/1.0.x/routing/#built-in-converters

This need was identified in #65 but I never took the time to dig into it.

deytao:
Thanks for your answer, I find a way to do the same by using the `location="path"` and the validator `AnyOf`.
It's not perfect but it does the job in the mean time.
I guess I can close this PR if you'll address it later.

lafrech:
All right. I started working on it but I'm not done yet.

Issue #69, opened 2019-05-07
swagger-ui: expose all parameters


lafrech:
See https://github.com/marshmallow-code/flask-smorest/pull/67#issuecomment-489207053.

> Maybe we could add the whole mapping from https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/configuration.md.
> 
> Care should be taken about retro-compatibility. I don't use swagger-ui and I don't know what happens when feeding an old version with a more recent parameter.

lafrech:
See what was done in https://github.com/sveint/flask-swagger-ui/blob/master/flask_swagger_ui/templates/index.template.html.

lafrech:
Closed by #171.

PR #171, opened 2020-07-07
Adding support for OPENAPI_SWAGGER_UI_CONFIG

joshua-harrison-2011:
This is patterned on https://github.com/sveint/flask-swagger-ui/blob/master/flask_swagger_ui/templates/index.template.html and is intended to address https://github.com/marshmallow-code/flask-smorest/issues/69.  I think it could also cover the use case in https://github.com/marshmallow-code/flask-smorest/pull/67.

### DIFF
--- a/docs/openapi.rst
+++ b/docs/openapi.rst
@@ -303,14 +303,15 @@ page using the JS script from the script URL.
 
    Default: ``None``
 
-.. describe:: OPENAPI_SWAGGER_UI_SUPPORTED_SUBMIT_METHODS
+.. describe:: OPENAPI_SWAGGER_UI_CONFIG
 
-   List of methods for which the '*Try it out!*' feature is enabled. Should be a
-   list of lowercase HTTP methods.
+   Dictionary representing Swagger UI configuration options.  See `Swagger UI Configuration`_ for available options.
+   All JSON serializable options are supported.
 
-   Passing an empty list disables the feature globally.
+   Examples:
+      * ``{'deepLinking': True, 'supportedSubmitMethods': ['get', 'post']}``
 
-   Default: ``['get', 'put', 'post', 'delete', 'options', 'head', 'patch', 'trace']``
+   Default: ``{}``
 
 Here's an example application configuration using both ReDoc and Swagger UI:
 
@@ -327,6 +328,7 @@ Here's an example application configuration using both ReDoc and Swagger UI:
 
 .. _ReDoc: https://github.com/Rebilly/ReDoc
 .. _Swagger UI: https://swagger.io/tools/swagger-ui/
+.. _Swagger UI Configuration: https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/
 
 Write OpenAPI Documentation File
 --------------------------------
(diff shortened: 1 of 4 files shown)


lafrech:
Thank you so much for tackling this.

Don't bother with backwards compatibility. We can drop `OPENAPI_SWAGGER_UI_SUPPORTED_SUBMIT_METHODS`.

I'll try to review more thoroughly when I get more time.

joshua-harrison-2011:
Thanks @lafrech.  I've removed the references to `OPENAPI_SWAGGER_UI_SUPPORTED_SUBMIT_METHODS`.

lafrech:
Thanks.

Should we add an example to the docs to make it clear to users camel-case options are expected? (We could use `supportedSubmitMethods` as example.)

Should we remove parameters hardcoded in the template, like `deepLinking` and `layout`? I must have copied this from an example a while ago, but we should stick to swagger-ui defaults as much as possible. (We can use `deepLinking` as example instead of `supportedSubmitMethods`, or maybe both.)

Can you comment on `safe` (in flask-swagger-ui) vs. `tojson` (in your code)?

joshua-harrison-2011:
Yes, I'll add a detailed example and remove `deepLinking` and `layout`.  

flask-swagger-ui doesn't use `tojson` because they are passing in a string in as config instead of the config object: https://github.com/sveint/flask-swagger-ui/blob/master/flask_swagger_ui/flask_swagger_ui.py#L35.  It seemed clearer to me to pass the object in and let the templating handle it, but I don't feel strongly about it.

I'll add `safe` back in - `var override_config = {{ swagger_ui_config | tojson | safe }};`.   I removed it when I thought it had to be `tojson` or `safe`, but forgot to add it back when I learned they could both be applied.

lafrech:
> flask-swagger-ui doesn't use tojson because they are passing in a string in as config instead of the config object

Oh right. They're passing a string already. TBH, I didn't search at all.

> It seemed clearer to me to pass the object in and let the templating handle it

Absolutely.

> I'll add `safe` back

According to [the docs](https://jinja.palletsprojects.com/en/2.11.x/templates/#tojson), `tojson` returns a safe result, so an additional `safe` shouldn't be needed IIUC. I may be totally mistaken. I have little experience with Jinja.

joshua-harrison-2011:
I agree that `safe` isn't necessary.  I missed that in the docs.  I added an example and removed the `deepLinking` and `layout` configuration parameters, but left `safe` out.

docs/openapi.rst
```diff
@@ -303,14 +303,15 @@ page using the JS script from the script URL.
 
    Default: ``None``
 
-.. describe:: OPENAPI_SWAGGER_UI_SUPPORTED_SUBMIT_METHODS
+.. describe:: OPENAPI_SWAGGER_UI_CONFIG
 
-   List of methods for which the '*Try it out!*' feature is enabled. Should be a
-   list of lowercase HTTP methods.
+   Dictionary representing Swagger UI configuration options.  See `Swagger UI Configuration`_ for available options.
+   All JSON serializable options are supported.
```
lafrech, line 309: Are there options that are not JSON serializable?

joshua-harrison-2011: That message is vague.  I added it because something similar was in the Readme for flask-swagger-ui.

There are several (9, I think) options that are defined as Javascript functions and wouldn't be representable in this config structure.

I could change the message to something like: `Function options are not supported.`.  Or maybe just get rid of it?

lafrech: Let's leave it as it is. I'm fine with your wording.

PR #183, opened 2020-08-27
Document int as int64 and float as double

lafrech:
Based on #182.

Rationale explained in https://github.com/marshmallow-code/apispec/pull/595.

Waiting for https://github.com/marshmallow-code/apispec/pull/595.


lafrech:
Replaced by #188.

PR #195, opened 2020-10-19
Support Python 3.9 and use Azure for tests

lafrech:


lafrech:
Travis does not support 3.9 right now. At least not this way.

Maybe it's time to move to Azure like the other marshmallow repos.

lafrech:
There's a little regression with the test coverage (https://github.com/asottile/azure-pipeline-templates/issues/36), but it is more of a theoretical issue right now as we don't have code conditioned by dependency or Python version.

OTOH, Travis config must be broken because we don't get the test results in the PRs and other marshmallow libs use Azure, so let's move on.

PR #208, opened 2020-12-23
Only add referenced responses to the spec

### LINKED ISSUE #206: Unreferenced responses appear in resulting OpenApi JSON
Hello!

I'm moving here from flask-rest-api and when rendering the my open api JSON file using the endpoint, i now have a LOT of responses... all possible ones if I believe the code.
All my endpoints are only referencing
```
#/responses/UNPROCESSABLE_ENTITY
#/responses/DEFAULT_ERROR
```
Added by flask-smorest it self.

Why would there be more than those 2 response in my /openapi.json output?

My versions, currently rendering with OPENAPI_VERSION=2.0
```
flask-smorest==0.26.0
apispec==4.0.0
marshmallow==3.10.0
``` 

Thank you!

lindycoder:
The process of adding responses dynamically to endpoints is relying on
adding the HTTP Status name as the response instead of an dict
referencing a schema.  This causes marsmallow to interpret it as a
reference so the reference has to exist.

Instead of adding all the possible responses to the spec, only the
responses referenced in any endpoints are added when necessary.  This
reduces noise in the spec while keeping the feature intact.

Fixes #206

==========

Good day sir,

What do you think of this approach?

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -17,3 +17,4 @@ Contributors (chronological)
 - Jón Bjarnason `@nonnib <https://github.com/nonnib>`_
 - Igor Davydenko `@playpauseandstop <https://github.com/playpauseandstop>`_
 - Joshua Harrison `@joshua-harrison-2011 <https://github.com/joshua-harrison-2011>`_
+- Martin Roy `@lindycoder <https://github.com/lindycoder>`_
--- a/flask_smorest/__init__.py
+++ b/flask_smorest/__init__.py
@@ -70,9 +70,6 @@ class Api(APISpecMixin, ErrorHandlerMixin):
         # Register error handlers
         self._register_error_handlers()
 
-        # Register responses
-        self._register_responses()
-
     def register_blueprint(self, blp, **options):
         """Register a blueprint in the application
 
--- a/flask_smorest/spec/__init__.py
+++ b/flask_smorest/spec/__init__.py
@@ -1,28 +1,23 @@
 """API specification using OpenAPI"""
 import json
-import http
 
 import flask
-from apispec import BasePlugin
 from flask import current_app
 import click
 import apispec
 from apispec.ext.marshmallow import MarshmallowPlugin
 
 from flask_smorest.exceptions import MissingAPIParameterError
-from flask_smorest.utils import prepare_response
-from .plugins import FlaskPlugin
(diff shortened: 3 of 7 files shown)


lafrech:
Thank you for your contribution. This PR is high quality. You really took the time to understand the internals of the lib, even of the tests. I appreciate.

I agree with the principle. Clever use of a Plugin to handle this.

The test coverage seems good.

I'm not sure the case where the same response is passed twice is explicitly tested. I'm thinking of this specific condition:

```py
    and response not in self._registered 
```

I shall double-check.

I'll be making a few changes (names, formatting), mostly nitpicks and personal preference, but this is pretty good.

Thanks.

lindycoder:
Glad you like it!

For the case of multiple response, if you take this out i see 15 failing tests with errors such as 
```
Another response with name "UNPROCESSABLE_ENTITY" is already registered
```
But you are right there is not a direct unit test for this case, only coverage from other tests, would like me to added a test specifically for this case in `TestRefResponsesPlugin`?

Also feel free to make your few changes !

Merci!

lafrech:
> would like me to added a test specifically for this case in TestRefResponsesPlugin?

Sure, if you have the time. Otherwise I'll do it.

lindycoder:
Sure, i'll do it right away!

lindycoder:
There you go, I took the liberty of actively testing all methods/response are iterated over and we don't blindly return after the first add_to_spec ;)

lafrech:
Thanks. Please also add yourself to AUTHORS.rst.

lafrech:
I rebased on master and added a rework commit.

I didn't manage to push to your branch. Not sure I'm even allowed to force push.

https://github.com/marshmallow-code/flask-smorest/compare/minimal_responses

Feedback welcome.

lindycoder:
I rebased and cherry picked your commit.

lafrech:
I'm tempted to change

```py
                         and response not in self._registered
```

into

```py
                         and response not in self.spec.components._responses
```

to allow users to set their own definitions before the default one is set.

(Using a private attribute but oh well...)

lindycoder:
The current master branch fits my needs perfectly, so if the behaviour doesn't change, I don't mind :)

lafrech:
No, it wouldn't be a breaking change.

I'm considering making the attribute public in apispec first.

lafrech:
> I'm considering making the attribute public in apispec first.

Done.

PR #210, opened 2020-12-23
Add a flag to disable default_errors

### LINKED ISSUE #209: Enhancement: Disable adding default_error to all responses
Hello! :)

So right now as soon as a `@blp.response()` is specified, a `default` response is added to cover all uncovered responses code.

The personal issue i have with this is that i may not be documenting all responses, such as unexpected 500s and if they happens, there's no way they will respect the default schema, which is not at all enforced anyway.

So I would rather not have this ambiguous claim that my responses respect this schema.

On a related note, if an endpoint only has `@blp.arguments()` and no `@blp.response()`, the default is not added, which in this case is probably a bug if the default is always expected to be there.

I suggest a flag to make adding default error response optional.

This may relate to #201

What do you think?

lindycoder:
Specified on the blueprint, this will skip adding default_error
responses automatically to endpoints, leaving the consumer to manually
manage those defaults.

Fixes #209 

========

It's me again!
Tell me what you think of this approach, I originally wanted to make this a flask app config, such as the `OPENAPI_*` configs, but endpoints are registered on the blueprint prior to initializing the flask app.

Having this on the blueprint level also allows consumers to subclass Blueprint and make it a default False if they have several blueprints.

### DIFF
--- a/flask_smorest/blueprint.py
+++ b/flask_smorest/blueprint.py
@@ -69,6 +69,7 @@ class Blueprint(
     def __init__(self, *args, **kwargs):
 
         self.description = kwargs.pop('description', '')
+        self.auto_default_error = kwargs.pop('auto_default_error', True)
 
         super().__init__(*args, **kwargs)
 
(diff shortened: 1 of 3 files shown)


lindycoder:
replaced by #222

PR #222, opened 2021-01-29
Make DEFAULT_ERROR optional (opt-out)

### LINKED ISSUE #209: Enhancement: Disable adding default_error to all responses

lafrech:
Replaces #210.
Closes  #209.

This PR

- makes the feature adding a default error optional (default: feature on)
- allows to choose the name of the default error in the spec (default: `"DEFAULT_ERROR"`, `None` disables the feature )
- ensures the default error is added to all resources (all functions decorated with `@route`, not only those decorated with `@response` (bug reported in #209)

As usual in flask-smorest, this is a class attribute. I deliberately didn't add an init parameter or a Flask app parameter for simplicity. I think the default is sensible and it is easy enough to subclass.

@lindycoder, what do you think about this implementation?

@lindycoder
> The personal issue i have with this is that i may not be documenting all responses, such as unexpected 500s and if they happens, there's no way they will respect the default schema, which is not at all enforced anyway.

My understanding is that the feature exists in OpenAPI spec as a shortcut to document similar responses as explained in https://swagger.io/docs/specification/describing-responses/. I added this default response in the docs because flask-smorest formats all errors with this structure, including 500 errors resulting from exceptions in the app. I don't think a client would blame the server for not complying on all server errors, like an error occurring at apache level. I mean there can always be a point were the server is so crashed it can't provide reliable results at all. So I'm not worried about this.

I'd be more worried about people adding custom decorators or decorators from other libs to implement features like authentication, rate limiting, etc. that wouldn't return a compliant error format. When I do so, I wrap the decorator to comply and to auto-document itself.

------------------------------------------------------

I'd like to release this along with #208.

This PR allows to go on with the #208 rework, declare the default error response on init and simplify `ResponseReferencesPlugin`.

### DIFF
--- a/flask_smorest/__init__.py
+++ b/flask_smorest/__init__.py
@@ -84,7 +84,7 @@ class Api(APISpecMixin, ErrorHandlerMixin):
         self._app.register_blueprint(blp, **options)
 
         # Register views in API documentation for this resource
-        blp.register_views_in_doc(self._app, self.spec)
+        blp.register_views_in_doc(self, self._app, self.spec)
 
         # Add tag relative to this resource to the global tag list
         self.spec.tag({'name': blp.name, 'description': blp.description})
--- a/flask_smorest/blueprint.py
+++ b/flask_smorest/blueprint.py
@@ -162,7 +162,7 @@ class Blueprint(
         # Store parameters doc info from route decorator
         endpoint_doc_info['parameters'] = parameters
 
-    def register_views_in_doc(self, app, spec):
+    def register_views_in_doc(self, api, app, spec):
         """Register views information in documentation
 
         If a schema in a parameter or a response appears in the spec
@@ -186,6 +186,7 @@ class Blueprint(
                     operation_doc = func(
                         operation_doc,
                         operation_doc_info,
+                        api=api,
                         app=app,
                         spec=spec,
                         method=method_l
--- a/flask_smorest/response.py
+++ b/flask_smorest/response.py
@@ -102,9 +102,6 @@ class ResponseMixin:
 
                 return resp
 
-            # Document default error response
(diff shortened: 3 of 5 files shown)


lindycoder:
HI @lafrech , sorry I took some time to answer.

To be honest, I'm not really a big fan of subclassing, but it's better than nothing I can live with it.

Perhaps you could add a test to show how the feature is used?  (Also to test the feature ;) )

lafrech:
Yes, I realized afterwards that I forgot to add a test. But then I forgot, so thanks for the reminder.

Regarding subclassing, this is consistent with the other options.

I understand it might look heavier to implement. My modest experience with the whole marshmallow / flask-smorest stack taught me that sooner or later I'll be willing to customize things so now I recommend to subclass right away (as demonstrated in [this outdated example I should update](https://github.com/lafrech/flask-smorest-sqlalchemy-example/blob/master/team_manager/extensions/api/__init__.py)).

lafrech:
> This PR allows to go on with the #208 rework, declare the default error response on init and simplify ResponseReferencesPlugin.

Scratch that. Not as nice as I thought. Not worth it.

PR #223, opened 2021-02-06
Allow manually registered HTTPStatus response

lafrech:
Allows a user to manually define responses overriding automatic HTTPStatus responses.

flask-smorest automatically defines responses when a HTTPStatus name is passed in an operation. This PR ensures that if such a response is already defined, it won't conflict (it won't raise) and the automatic definition will be skipped.

See #208 for background.

Not a breaking change except it bumps apispec min version to 4.2.0.


lafrech:
This would be reworked if we implemented #236. But we can merge it anyway, it doesn't matter. Just waiting for a "major" (0.x) version to bump apispec version.

lafrech:
Obsolete.

This can be achieved in user code by overriding `Api._register_responses`.

PR #227, opened 2021-02-09
Use werkzeug.Response, not BaseResponse

lafrech:
Shouldn't hurt with werkzeug < 2.

BaseResponse is deprecated in werkzeug 2.

### DIFF
--- a/flask_smorest/response.py
+++ b/flask_smorest/response.py
@@ -4,7 +4,7 @@ from copy import deepcopy
 from functools import wraps
 import http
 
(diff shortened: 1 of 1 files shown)


lafrech:
On second thought, this would be a breaking change if someone specified a base class that is a BaseResponse but not a Reponse. Unlikely case but let's wait for werkzeug 2 to ship this.

lafrech:
Let's wait until Werkzeug 2 and Flask 2 are out and do this in #243.

PR #237, opened 2021-04-06
added support for 'servers' configuration on OpenAPI v3

aprilahijriyan:
added support for 'servers' configuration on OpenAPI v3 based on `APPLICATION_ROOT`.
This is for applications wrapped in `DispatcherMiddleware`.
Already tested using this https://github.com/zemfrog/zemfrog

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+.vscode
+demo/
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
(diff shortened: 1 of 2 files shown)


aprilahijriyan:
Hi @lafrech, Let me know if anything is missing.

lafrech:
Hi @aprilahijriyan.

I'm willing to go the opposite way. I think it can't reasonably and reliably be manual.

`APPLICATION_ROOT` is not the right config param to check. It is meant for cookies in Flask. It is generally not needed for a REST API. I don't set it in my apps. Using this in the first place was a mistake.

I'll remove the whole feature and let the user handle it rather than let him think it all works fine automagically.

BTW, this is how it is done in [APIFlask](https://github.com/greyli/apiflask). `servers` is a user config param.

Sidenote: Don't add stuff related to you dev environment in projects .gitignore. This is what your global .gitignore is for. See https://docs.github.com/en/github/getting-started-with-github/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer.

PR #241, opened 2021-04-19
Fix the type of examples in docstring

greyli:
The `examples` argument for the request/response body should be a dict ([ref](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.2.md#request-body-examples)).


lafrech:
Right. Thanks.

Care to add yourself to AUTHORS.rst?

greyli:
Added. :)

lafrech:
Thanks!

PR #243, opened 2021-05-03
Move to werkzeug 2 and flask 2

lafrech:
When those are ready.

### DIFF
--- a/flask_smorest/response.py
+++ b/flask_smorest/response.py
@@ -4,7 +4,7 @@ from copy import deepcopy
 from functools import wraps
 import http
 
-from werkzeug.wrappers import BaseResponse
+from werkzeug import Response
 from flask import jsonify
 
 from .utils import (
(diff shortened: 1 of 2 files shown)


flask_smorest/response.py
```diff
@@ -4,7 +4,7 @@
 from functools import wraps
 import http
 
-from werkzeug.wrappers import BaseResponse
+from werkzeug import Response
```
AbdealiLoKo, line 7: Maybe just doing both would be best ? That way we can support `'werkzeug>=0.15,<3'` and `flask>=2.0,<3`
For example:
```python
from werkzeug.wrappers import BaseResponse
respons_classes = (BaseResponse,)
try:
    from werkzeug import Response
    respons_classes += (Response,)
except ImportError:
    pass
...
isinstance(result_raw, respons_classes)

```

lafrech: Sure.

I just find it simpler to drop older versions of dependencies while we're still in 0.x here. Keeping Flask + Werkzeug v1 compatibility would require to actually test on them to ensure nothing gets broken in the future.

I would see it differently if the migration to Werkzeug/Flask v2 was difficult, but the people at pallets have done a nice job of ensuring the transition is smooth and I don't expect too many difficulties.

Unless someone feels strongly against it, I'm willing to drop Flask/Werkzeug < 2.

AbdealiLoKo: I'm good dropping support for older versions

lafrech:
I went through the changelogs again and I think the transition should be easy so there's no real incentive to support older versions.

Moving forward with this.

PR #254, opened 2021-06-14
Generate header parameter and responses in API when ETag is set

### LINKED ISSUE #162: Document ETag header
When using etag decorator, etag related errors are documented, but the etag header is not.

TimDiekmann:
Closes #162

### DIFF
--- a/flask_smorest/etag.py
+++ b/flask_smorest/etag.py
@@ -250,9 +250,30 @@ class EtagMixin:
             method_u = method.upper()
             if method_u in self.METHODS_CHECKING_NOT_MODIFIED:
                 responses[304] = http.HTTPStatus(304).name
+                doc.setdefault("parameters", []).append({
+                    'name': 'If-None-Match',
+                    'in': 'header',
+                    'description': 'Tag to check against',
+                    'schema': {'type': 'string'}})
             if method_u in self.METHODS_NEEDING_CHECK_ETAG:
                 responses[412] = http.HTTPStatus(412).name
                 responses[428] = http.HTTPStatus(428).name
+                doc.setdefault("parameters", []).append({
+                    'name': 'If-Match',
+                    'in': 'header',
+                    'required': True,
+                    'description': 'Tag to check against',
+                    'schema': {'type': 'string'}})
+            if method_u in self.METHODS_ALLOWING_SET_ETAG:
+                success_status_code = doc_info.get('success_status_code')
+                if success_status_code is not None:
+                    header = {
+                        'description': 'Tag for the returned entry',
+                        'schema': {'type': 'string'}
+                    }
+                    doc['responses'][success_status_code] \
+                        .setdefault('headers', {})['ETag'] = header
+
             if responses:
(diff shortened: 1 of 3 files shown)


flask_smorest/etag.py
```diff
@@ -250,9 +250,30 @@ def _prepare_etag_doc(self, doc, doc_info, *, app, method, **kwargs):
...
                 responses[428] = http.HTTPStatus(428).name
+                doc.setdefault("parameters", []).append({
+                    'name': 'If-Match',
+                    'in': 'header',
+                    'required': True,
+                    'description': 'Tag to check against',
+                    'schema': {'type': 'string'}})
+            if method_u in self.METHODS_ALLOWING_SET_ETAG:
+                success_status_code = doc_info.get('success_status_code')
+                if success_status_code is not None:
```
TimDiekmann, line 269: Branch coverage fails here. Do you want me to add a new test case without `@blp.response(204)`?

TimDiekmann, line 269: Branch coverage fails here. Do you want me to add a new test case without `@blp.response(204)`?

lafrech: This is a good catch from the coverage checker. I didn't see that when reading the code. The ETag header is correctly documented only if `@blp.response` is used because otherwise there's no way to tell which status code corresponds to the "normal" path.

If the user doesn't use `@blp.response`, because he creates the response manually, the ETag header can't be auto-documented.

This is a case where the user should use `@blp.alt_response`. Perhaps I could make this clearer in the docs: `@blp.alt_response` is not only for responses that are not the "normal flow" (like exceptions,...) but also responses that are returned as `Response` object thus requiring no serialization work from flask-smorest. Anyway, we could add an argument to allow the user to specify that an "alt_response" is in fact a "success" one.

Thinking of it, there may be several "success" responses. The point is that we exit the view function without exception and go through the decorator wrappers. Perhaps the whole "success" thing should be redesigned. At least we could change it to be a list of "successful status codes", meaning "non aborted responses". And perhaps the argument to `@blp.alt_response` could be `aborted` or `exception`, defaulting to `True`.

I need to think this through. Feedback about the feature and argument name ideas welcome.

Meanwhile, yes, you could add a case without `@blp.response` showing the ETag is not documented.

If we go through the idea above, we'll update the test to use `@blp.alt_response` with and without success. And perhaps warn if the user uses ETag without `@blp.response` or `@blp.alt_response` as this makes the docs incomplete.

Thanks for your contribution. You've taken the time to understand the internals and this is a good PR.

Please also add yourself to AUTHORS.rst.

TimDiekmann: Adding multiple success responses sounds reasonable. Personally, I didn't need more than one success response, but that's just a simple REST API without any fancy features or logic. In any case, it would be nice to be able to mark a response as a sucess.

I parametrized `has_response` and added me to *AUTHORS.rst*, many thanks for that!

TimDiekmann:
"100.00% of diff hit" 😁

lafrech (approved):
Thanks!

PR #255, opened 2021-06-18
Use click.echo() for spec output commands

greyli:
The `click.echo` function is a more proper way to output messages to stdout in `flask openapi print` command, as it states in [the docs](https://click.palletsprojects.com/en/8.0.x/api/#click.echo):

> Print a message and newline to stdout or a file. This should be used instead of print() because it provides better support for different data, files, and environments.

For consistency, this PR also change the `flask openapi write` command to use `click.echo` to write the file.


lafrech (approved):
Thanks. I didn't know that.

Mind adding yourself to AUTHORS file?

greyli:
> Mind adding yourself to AUTHORS file?

I'm already in it :p

Issue #252, opened 2021-06-10
Multiple separate apis per-application ?


ddorian:
Looks like some configs like `OPENAPI_URL_PREFIX` are fixed to the application and this prevents the ability to have multiple apis for 1 app. The simplest case I'd think 1 Api() per major-version. So you'd want a completely separate api between /v1/ and /v2/.

lafrech:
To support several API versions, we run multiple versions of the code at different URLs.

I never tried to have multiple API versions in the same application version. I don't know how I'd achieve this with flask-smorest. It is not only about the documentation. How do you separate the views from v1 and v2?

I guess you could do it by nesting blueprints in a v1 and a v2 blueprint, using the Flask 2 blueprint nesting feature. But limiting duplication would require reusing blueprints, therefore parametrizing them for the v1 vs. v2 differences. And I'm not sure about the support of nested blueprints in flask-smorest. This is a new Flask feature I haven't tested.

As of today, multiple APIs per app is not supported, and I don't deem the benefit/cost high enough to try to do it.

Anyone feeling differently, please comment.

ddorian:
> I don't know how I'd achieve this with flask-smorest.

Change all the settings to per-app to be per-api object. You have 1 swagger-ui,redoc,etc per-api instead for the whole application. 

> How do you separate the views from v1 and v2?

You just need a way to connect a blueprint to an Api() and that's it. Think everything that currently is global, gets connected to Api(). When you register a blueprint, you also register it with an Api() and that's how you can distinguish things from each other.

lafrech:
Oh, right. 1 Flask `app` with multiple flask-smorest `Api`s.

How would you prefix the parameters to distinguish the `Api`s in the app config context? The `Api` object has no `name` attribute. Shall we add one?

ddorian:
> The Api object has no name attribute. Shall we add one?

Yes, `name` can be used to distinguish them.


> How would you prefix the parameters to distinguish the Apis in the app config context?

Instead of doing:
```python
app.config["OPENAPI_URL_PREFIX"] = ""
```
I personally would use my own keys:
```python
Api(OPENAPI_URL_PREFIX=app.config["API_V1_URL_PREFIX"])
```

If you want to make it dynamic, then something like:
```
Api(api_name="api_v1")
app.config["SMOREST"] = {"api_v1": {"OPENAPI_URL_PREFIX": ""}}
```
Another way would be to concat `SMOREST` with `Api().name` and config_name:
```
[...]

PR #259, opened 2021-06-30
allow DocBlueprintMixin to be configured using kwargs

sjhewitt:
This makes it possible to create 2 Api instances with the UI for the same app


lafrech:
Thank you for contributing.

Closing this because I'd rather take another approach. See #252.

PR #262, opened 2021-07-25
Fix rapidoc config when passing values with whitespaces

TimDiekmann:
When passing config values, like `api-key-value="Bearer TOKEN"`, rapidoc will omit everything behind the whitespace.


lafrech:
Thanks!

PR #272, opened 2021-08-17
Add Blueprint.add_url_rule

### LINKED ISSUE #269: Unable to use add_url_rule() directly
Currently, in one of my projects we use `blueprint.add_url_rule()` and we are trying to move towards using smorest.

When we tried this - it seems like the documentation is not getting updated when we use `add_url_rule`.
On digging further, I see that `blueprint.route()` is doing some extra logic which is not happening if I directly try to add the rule

https://github.com/marshmallow-code/flask-smorest/blob/master/flask_smorest/blueprint.py#L127

This is a blocker for us - as we have our own custom classes that we are using - kind of like the Django viewsets.
So, basically we cannot decorate the class.
For example: We have a class `MyClass` and we create 2 views from it:
 - `MyClass.create_view(action={'get': 'list' , 'post': create'})` for `/`
 - and `MyClass.create_view(action={'post': 'edit', 'delete': 'delete'})` for `/<id>`

It would be good to have a function to do this logic instead of just a decorator

lafrech:
Override `Blueprint.add_url_rule` to allow the use of `add_url_rule`, not only the `route` decorated form.

Closes #269.

### DIFF
--- a/flask_smorest/blueprint.py
+++ b/flask_smorest/blueprint.py
@@ -90,42 +90,63 @@ class Blueprint(
             self._prepare_etag_doc,
         ]
 
-    def route(self, rule, *, parameters=None, **options):
-        """Decorator to register url rule in application
+    def add_url_rule(
+        self,
+        rule,
+        endpoint=None,
+        view_func=None,
+        provide_automatic_options=None,
+        *,
+        parameters=None,
+        **options,
+    ):
+        """Register url rule in application
 
         Also stores doc info for later registration
 
-        Use this to decorate a :class:`MethodView <flask.views.MethodView>` or
+        Use this to register a :class:`MethodView <flask.views.MethodView>` or
         a resource function.
 
         :param str rule: URL rule as string.
         :param str endpoint: Endpoint for the registered URL rule (defaults
             to function name).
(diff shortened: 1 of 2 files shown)


flask_smorest/blueprint.py
```diff
@@ -90,42 +90,63 @@ def __init__(self, *args, **kwargs):
...
+            func = view_func
 
-            # Add URL rule in Flask and store endpoint documentation
-            self.add_url_rule(rule, endpoint, view_func, **options)
-            self._store_endpoint_docs(endpoint, func, parameters, **options)
+        # Add URL rule in Flask and store endpoint documentation
+        super().add_url_rule(rule, endpoint, func, **options)
+        self._store_endpoint_docs(endpoint, view_func, parameters, **options)
 
+    def route(self, rule, *, parameters=None, **options):
```
AbdealiLoKo, line 141: This function does not seem to be required anymore - as it is an exact copy of the parent - flask.scaffold.Scaffold.route()

lafrech: Yeah, I noticed that. I think I left it so that it appears in the documentation, exposing "parameters" parameter and the docstring.

Issue #110, opened 2019-11-26
Undocumented routes


mvanderlee:
I have a use-case where I want to have an undocumented internal route, while still using the arguments and response parsing. 

Perhaps something like 
`@blp.route('/hidden-api', document=False)`

lafrech:
I understand the feature but I'm not sure it is worth the added API surface.

It should be easy to achieve in user code.

Something like

```py
class MyBlueprint(Blueprint):
    def route(self, rule, *, parameters=None, document=True, **options):
        if document:
            return super().route(rule, parameters=parameters, document=document, **options)
        # Copy else case from Flask Blueprint
        else:
            def decorator(f):
                endpoint = options.pop("endpoint", f.__name__)
                self.add_url_rule(rule, endpoint, f, **options)
                return f
            return decorator      
```

Would this work for you?

I'm not definitively and strongly opposed to the feature. Just unsure.

mvanderlee:
This certainly works for me.

0x78f1935:
I would love to see this feature integrated. 

Edit:

Here is an updated version!

```py
# encoding: utf-8
"""
Blueprint adapter
-----------------
Flask-Smorest is not supporting hidden endpoints.
This class follows the following suggestion: https://github.com/marshmallow-code/flask-smorest/issues/110
"""
from flask_smorest import Blueprint
from flask.views import MethodViewType


class EnhancedBlueprint(Blueprint):
    def route(self, rule, *, parameters=None, render=True, **options):
        if render: return super().route(rule, parameters=parameters, **options)
        else: # Hide documentation and swagger UI elements
            def decorator(func):

                # By default, endpoint name is function name
                endpoint = options.pop('endpoint', func.__name__)

                # Prevent registering several times the same endpoint
                # by silently renaming the endpoint in case of collision
                if endpoint in self._endpoints:
                    endpoint = '{}_{}'.format(endpoint, len(self._endpoints))
                self._endpoints.append(endpoint)

                if isinstance(func, MethodViewType):
                    view_func = func.as_view(endpoint)
                else:
                    view_func = func

                # Add URL rule in Flask and store endpoint documentation
                self.add_url_rule(rule, endpoint, view_func, **options)
                # self._store_endpoint_docs(endpoint, func, parameters, **options)

                return func

            return decorator
```

Edit 2: I created a Pull Request

mvanderlee:
This solutions no longer works due to internal updates in flask-smorest.

PR #276, opened 2021-08-24
Added parameter to hide documentation in the swagger UI

0x78f1935:
Came across the following issue: https://github.com/marshmallow-code/flask-smorest/issues/110
Decided to make a PR for it.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -22,3 +22,4 @@ Contributors (chronological)
 - Stephen Rosen `@sirosen <https://github.com/sirosen>`_
 - Grey Li `@greyli <https://github.com/greyli>`_
 - Tim Diekmann `@TimDiekmann <https://github.com/TimDiekmann>`_
+- 0x78f1935 `@0x78f1935 <https://github.com/0x78f1935>`_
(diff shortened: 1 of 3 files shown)


lafrech:
Hi. Thanks for contributing.

Not sure about the name of the parameter. `document` sounds too generic. I don't find `render` very expressive. `hide_from_doc` would be expressive. I'm open to suggestions.

Regarding the implementation, propagate the parameter to `add_url_route` and make things happen there.

0x78f1935:
Sorry I was sleeping or something, latest version should work 👍

0x78f1935:
Any news on this? @lafrech

lafrech (changes requested):
I can't find great a parameter name. I think `hidden_from_api_spec` would be better. I shall make this consistent across the lib, but the JSON that is produced is more a specification than a documentation (if I understand [this](https://swagger.io/resources/articles/difference-between-api-documentation-specification/) correctly).

Add unit tests, squash the commits and you're done.

Also, please add your name to AUTHORS file (can be in a separate commit).

PR #277, opened 2021-09-03
Added new parameter in Route decorator 'hidden_from_api_spec';

0x78f1935:
* Added new parameter in Route decorator 'hidden_from_api_spec'
* Added related unittest
* Added missing required libraries for unittests in dev-requirements.txt
* Updated AUTHORS.rst

See https://github.com/marshmallow-code/flask-smorest/pull/276

I never squashed myself and i Kinda screwed up the previous pr.
Re-forked the original library and applied the changes.

@lafrech Thank you for taking the time for those changes!

On another note, I've found some alternatives for ReDoc which I would like to implement. I might create a new PR for this if you are oke with this. See https://stackoverflow.com/questions/36634281/list-of-swagger-ui-alternatives

### DIFF
--- a/flask_smorest/blueprint.py
+++ b/flask_smorest/blueprint.py
@@ -99,6 +99,7 @@ class Blueprint(
         *,
         parameters=None,
         tags=None,
+        hidden_from_api_spec=False,
         **options,
     ):
         """Register url rule in application
@@ -116,6 +117,9 @@ class Blueprint(
             in this path, only used to document the resource.
         :param list tags: List of tags for the resource.
             If None, ``Blueprint`` name is used.
+        :param boolean hidden_from_api_spec: Don't document resource in API
+            spec file.
+            Default False.
         :param options: Options to be forwarded to the underlying
             :class:`werkzeug.routing.Rule <Rule>` object.
         """
@@ -139,10 +143,19 @@ class Blueprint(
 
         # Add URL rule in Flask and store endpoint documentation
         super().add_url_rule(rule, endpoint, func, **options)
-        self._store_endpoint_docs(
-            endpoint, view_func, parameters, tags, **options)
+        if not hidden_from_api_spec:
+            self._store_endpoint_docs(
+                endpoint, view_func, parameters, tags, **options)
 
-    def route(self, rule, *, parameters=None, tags=None, **options):
+    def route(
+            self,
+            rule,
+            *,
+            parameters=None,
+            tags=None,
+            hidden_from_api_spec=False,
+            **options
+            ):
         """Decorator to register view function in application and documentation
 
         Calls :meth:`add_url_rule <Blueprint.add_url_rule>`.
@@ -155,6 +168,7 @@ class Blueprint(
                 func,
                 parameters=parameters,
                 tags=tags,
+                hidden_from_api_spec=hidden_from_api_spec,
                 **options
             )
             return func
--- a/tests/test_blueprint.py
+++ b/tests/test_blueprint.py
@@ -570,6 +570,24 @@ class TestBlueprint:
         )
         assert spec["paths"]["/test/test_2/"]["get"]["tags"] == ["test", ]
 
+    def test_blueprint_route_hidden_from_api_spec(self, app):
+        """Check passing hidden_from_api_spec to route"""
+        blp = Blueprint('test', __name__, url_prefix='/test')
+
+        @blp.route('/test_1/', hidden_from_api_spec=True)
+        def test_1():
+            pass
+
+        @blp.route('/test_2/')
+        def test_2():
+            pass
+
+        api = Api(app)
+        api.register_blueprint(blp)
+        spec = api.spec.to_dict()
+        assert "/test/test_1/" not in spec["paths"]
+        assert "/test/test_2/" in spec["paths"]
+
     @pytest.mark.parametrize('openapi_version', ['2.0', '3.0.2'])
     def test_blueprint_response_schema(self, app, openapi_version, schemas):
         """Check response schema is correctly documented.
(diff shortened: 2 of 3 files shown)


0x78f1935:
@lafrech Any updates on this? Thanks in advance

flask_smorest/blueprint.py
```diff
@@ -116,6 +117,8 @@ def add_url_rule(
             in this path, only used to document the resource.
         :param list tags: List of tags for the resource.
             If None, ``Blueprint`` name is used.
+        :param boolean hidden_from_api_spec: Hide documentation in swagger UI.
+            Default False.
```
lafrech, line 122: Not only Swagger-UI.

```py
        :param boolean hidden_from_api_spec: Don't document resource in API
            spec file.
            Default False.
```

tests/test_blueprint.py
```diff
@@ -570,6 +570,23 @@ def test_2():
...
+            pass
+
+        @blp.route('/test_2/')
+        def test_2():
+            pass
+
+        api = Api(app)
+        api.register_blueprint(blp)
+        spec = api.spec.to_dict()
+        assert (len(spec["paths"]) == 1)
```
lafrech: I'd test this instead of the length.

```py
        assert "/test/test_1/" not in spec["paths"]
        assert "/test/test_2/" in spec["paths"]
```

lafrech (changes requested):
Looks good. Just minor comments about wordings.

------------------------------------------------

I don't think we need to modify dev-requirements. What I do is

```
pip install -e .
pip install -r dev-requirements
```

I shall pin the dependencies someday and activate dependabot, but I'll do it in another PR and with strict equalities.

flask_smorest/blueprint.py
0x78f1935: I've fixed this locally, how would you like me to commit it? I would like to avoid squashing commits if possible

tests/test_blueprint.py
0x78f1935: Ohh alrighty, learned something new. I leave the requirements file for what it is.

I applied the changes for the unittest locally, would you like me to commit it all in one commit or would you like me to make 2 commits. One for docstring and one for unittest?

lafrech: Ultimately, you should learn the magic of `rebase -i`. It comes in very handy to rewrite history.

For now, the easiest path for you if you're not comfortable with squasing is to amend your last commit then push force.

```
git add .
git commit --amend
git push -f
```

0x78f1935: Yes, you are right. I might actually just make a repository to play around with this. --amend was also new for me.

Thank you for your input, much appriciated!

I've committed the changes : )

lafrech (approved):

lafrech (changes requested):
Please undo the dev-requirements.txt change.

```
git reset HEAD~1
git co dev-requirements.txt
git add flask_smorest tests
git commit -m "Add hidden_from_api_spec parameter to route/add_url_rule"
git add . AUTHORS.rst
git commit -m "Add 0x78f1935 to AUTHORS.rst"
git push -f
```

0x78f1935:
Uhm before approving, perhaps it is better to change `hidden_from_api_spec` from boolean to a list.
An empty list represents False (default value).

The list itself can contain the following responses ['GET', 'POST', 'UPDATE', 'PUT', 'DELETE', etc etc].
Which should hide only the methods listed in `hidden_from_api_spec`.

This way you can hide endpoints such as
`@blp.route('/auth/oauth2/authorize', methods=['GET', 'POST'], hidden_from_api_spec=['GET'])` 

What do you think about this proposal?

0x78f1935:
I'm sorry but it takes pretty long before obtaining any answers, I really could use this feature. @lafrech any updates?

lafrech:
I'm reluctant to add API surface for a feature that I believe is a corner case. I'd be happier if this was easily achievable in user code. Unfortunately, I can't find a way to do that.

It would be easy if we could pass arbitrary kwargs to werkzeug's `add_url_rule`. Then the user would just have to override `_store_endpoint_docs` and get `hidden_from_api_spec`  from `**options`.

Since this is not feasible, I'm willing to add the feature to the core. I'd like to keep it simple but the method list is a legit case too and also wouldn't be easy to achieve without duplicating too much code.

You may propose an implementation.

`hidden_from_api_spec` may be `False` (default), `True` (hide all methods) or a list of methods to hide. (Don't use `[]` as default...)

Pass `hidden_from_api_spec` to `_store_endpoint_docs` and in there, at the beginning, use it to filter `self.HTTP_METHODS` (keep the order).

-----------------------

Regarding answer delays, please realize that most of the maintenance work in this organization is benevolent work. Adding features we need and keeping the lib up to date is part of my work, but working on corner-case features we won't be using is not and at the end of the day, I end up doing maintenance chores on my free time. I'm not complaining, just saying I can't guarantee answer delays. In this context, bug fixes get higher priority than such features.

It also takes time to answer because before letting someone code a PR, I try to be sure they are on the right track to avoid wasting their time.

Lastly, an important part of the work to have this lib running happens in webargs, apispec and marshmallow itself and may not be visible here.

If you really need a feature, you may override or fork until it gets added. Pinging maintainers generally doesn't help.

0x78f1935:
> I'm reluctant to add API surface for a feature that I believe is a corner case.

This makes total sense. I get why.

> If you really need a feature, you may override or fork until it gets added. Pinging maintainers generally doesn't help.

It's kinda that the "higher hierarchy" expected me to go after this PR since we are about to use flask_smorest in our production environment. I'm sorry if i came over as a little pushy. This is not my intention. I can make a workaround, so don't worry about it. The thing is that I end up with messy unnecessary duplicated code just like you described 😅 .

Thank you for your response, i can forward this to my project leader.

----

I'm almost done with finalizing our API, There are a few endpoints I would like to hide so I will come back to this.

> `hidden_from_api_spec` may be `False` (default), `True` (hide all methods) or a list of methods to hide. (Don't use `[]` as default...)
> 
> Pass `hidden_from_api_spec` to `_store_endpoint_docs` and in there, at the beginning, use it to filter `self.HTTP_METHODS` (keep the order).

I see what I can do, but those instructions are very clear. I let you know!

PR #305, opened 2021-12-13
Respect JSON_SORT_KEYS in _generate_etag

lafrech:
After all, there is no reason to force this here. Respecting the setting from Flask makes things more consistent with payload dump.

This change will affect users setting `JSON_SORT_KEYS` to `False` in that generated ETags will now differ if the order of the dict payload changes. This makes sense for a strong ETag.

There are more things I'd like to clean up in the ETag feature. We're doing too much in flask-smorest and we should rely more on Flask/Werkzeug. This is just an easy first step.

This could be considered a (minor) bug fix. And a (slightly) breaking change.

### DIFF
--- a/flask_smorest/etag.py
+++ b/flask_smorest/etag.py
@@ -141,9 +141,8 @@ class EtagMixin:
             raw_data = etag_schema.dump(etag_data)
         if extra_data:
             raw_data = (raw_data, extra_data)
-        # flask's json.dumps is needed here
-        # as vanilla json.dumps chokes on lazy_strings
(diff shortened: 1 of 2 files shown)


flask_smorest/etag.py
```diff
@@ -143,7 +143,7 @@ def _generate_etag(etag_data, etag_schema=None, extra_data=None):
             raw_data = (raw_data, extra_data)
         # flask's json.dumps is needed here
         # as vanilla json.dumps chokes on lazy_strings
```
lafrech: Update comment to mention other reasons (_dump_arg_defaults).

PR #312, opened 2022-01-08
Update error_handler.py to support description

juur:
Currently raising a werkzeug HttpException exception, e.g. BadRequest(description="<text>"), results in a JSON object returned omitting the description field, versus the non-smorest scenario where the body of the text/html response includes the description.

### DIFF
--- a/flask_smorest/error_handler.py
+++ b/flask_smorest/error_handler.py
@@ -54,6 +54,12 @@ class ErrorHandlerMixin:
         """
         headers = {}
         payload = {"code": error.code, "status": error.name}
+        
+        # To cover the example:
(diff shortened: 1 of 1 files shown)


juur:
Hi - any chance this can be merged?

lafrech:
The typical way of doing this is to use `abort` (from flask, enhanced by webargs).

I don't have a strong objection to supporting werkzeug exceptions. Doing so would require tests, docstring update and doc update.

But why not use

```py
    abort(400, message="<text>")
```

or

```py
    abort(http.HTTPStatus.BAD_REQUEST, message="<text>")
```

rather than

```py
    raise werkzeug.exceptions.BadRequest(description="<text>")
```

?

Issue #365, opened 2022-06-09
Flask commands to generate OpenAPI schema in yaml 


derlikh-smart:
Currently there are two Flask command:

1. `flask openapi print` to print OpenAPI schema to output
2. `flask openapi write <file>` to write OpenAPI schema to file

Both of them serialize OpenAPI schema to json.
It would be great to have an option to generate it to yaml as well.

lafrech:
I'm reluctant to introduce a dependency to yaml but if it was optional like in apispec, I guess I wouldn't object.

I won't do it myself, but people interested in this may discuss it here and propose an implementation.

derlikh-smart:
I wanted to implement it myself. But first of all it makes sense to create an issue to discuss it.
If it's OK, I'll work on PR.

lafrech:
Achieved in #367.

PR #366, opened 2022-06-09
Flask command generate schema yaml

derlikh-smart:
Implements #365

### DIFF
--- a/flask_smorest/spec/__init__.py
+++ b/flask_smorest/spec/__init__.py
@@ -358,7 +358,7 @@ def print_openapi_doc(format):
     """Print OpenAPI JSON document."""
     if format == "json":
         click.echo(json.dumps(_get_spec_dict(), indent=2))
-    elif format == "yaml":
+    else:  # format == "yaml"
         if HAS_PYYAML:
             click.echo(yaml.dump(_get_spec_dict()))
         else:
@@ -374,7 +374,7 @@ def write_openapi_doc(format, output_file):
     """Write OpenAPI JSON document to a file."""
     if format == "json":
         click.echo(json.dumps(_get_spec_dict(), indent=2), file=output_file)
-    elif format == "yaml":
+    else:  # format == "yaml"
         if HAS_PYYAML:
             yaml.dump(_get_spec_dict(), output_file)
         else:
--- a/flask_smorest/spec/__init__.py
+++ b/flask_smorest/spec/__init__.py
@@ -1,6 +1,7 @@
 """API specification using OpenAPI"""
 import json
 import http
+import sys
 
 import flask
 from flask import current_app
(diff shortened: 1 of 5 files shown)


flask_smorest/spec/__init__.py
```diff
@@ -340,16 +347,37 @@ def _register_pagination_header(self):
 openapi_cli = flask.cli.AppGroup("openapi", help="OpenAPI commands.")
 
 
+def _get_api():
+    return current_app.extensions["flask-smorest"]["ext_obj"]
+
```
lafrech, line 353: I wouldn't have bothered factoring this but this is merely a matter or personal preference.

lafrech (commented):
Thanks for this.

Do you think it could be possible to have a single print command with format as an optional argument defaulting to JSON (and likewise for write)?

This would make for an (arguably) nicer user interface. It might also help factorize code, although this is not my concern, here.

flask_smorest/spec/__init__.py
derlikh-smart: From my point of view when something (even a single line) is copy/pasted more then two times, it makes sense to extract it somehow.

In the original code it was exactly two times :-)

derlikh-smart: In original co

> Thanks for this.
> 
> Do you think it could be possible to have a single print command with format as an optional argument defaulting to JSON (and likewise for write)?
> 
> This would make for an (arguably) nicer user interface. It might also help factorize code, although this is not my concern, here.

I think it's a good idea! I didn't think about it. I will update the PR today or tomorrow.

Do you think it would be possible to release soon after the PR is merged, so I can use these commands?

lafrech: Can't make promises but I'll try to ship fast.

Regarding the duplicated line, it is just a dict access and it could be argued that the function call hinders readability a bit. But no strong objection. (In any case, I can always rework it someday if I really feel the need.)

tests/test_spec.py
```diff
@@ -409,3 +413,117 @@ def test_apispec_serve_spec_preserve_order(self, app):
...
+
+
+@pytest.fixture
+def flask_cli_runner(app):
+    return app.test_cli_runner()
+
+
+@pytest.fixture
+def temp_file(tmp_path):
+    file_path = tmp_path / "".join(
```
lafrech: Why not use `tmp_path` as shown in [pytest docs](https://docs.pytest.org/en/6.2.x/tmpdir.html)?

The file name doesn't have to be random.

```py
def test_create_file(tmp_path):
    output_file = tmp_path / "openapi_file"
```

lafrech (commented):
Good.

Wondering if the param name should be output or format.

Isn't -o often use for output file path in command line utilities? (I don't have many examples to back this up, just a feeling.)

tests/test_spec.py
derlikh-smart: You're right. I've set it to fixed file name.

derlikh-smart:
> Good.
> 
> Wondering if the param name should be output or format.
> 
> Isn't -o often use for output file path in command line utilities? (I don't have many examples to back this up, just a feeling.)

Good catch. Removed the short option.

tests/test_spec.py
```diff
@@ -409,3 +412,96 @@ def test_apispec_serve_spec_preserve_order(self, app):
...
+    return app.test_cli_runner()
+
+
+@pytest.fixture
+def temp_file(tmp_path):
+    file_path = tmp_path / "openapi_file"
+    yield str(file_path)
+    with contextlib.suppress(FileNotFoundError):
+        file_path.unlink()
+
```
lafrech, line 412: My point was that you could remove the fixture and use the pytest fixture directly in each test.

```py
def test_apispec_command_write(tmp_path):
    result = flask_cli_runner.invoke(args=cmd.split() + [tmp_path / "openapi_file"])
```

lafrech (commented):
Regarding the option name, it's not just about -u vs. --output.

I had the feeling that --format / -f could be less ambiguous.

Again, I don't have stats/evidence to back this up. Perhaps format is ambiguous as well. I just want to get it right the first time.

tests/test_spec.py
derlikh-smart: I don't like, when tests leave side effects after the execution. With fixture it's much easier to control it.

derlikh-smart:
> Regarding the option name, it's not just about -u vs. --output.
> 
> I had the feeling that --format / -f could be less ambiguous.
> 
> Again, I don't have stats/evidence to back this up. Perhaps format is ambiguous as well. I just want to get it right the first time.

I took the option name from [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-output-format.html).
But I agree, `-f,--format` is more clear. I'll update it.

lafrech:
I squashed and did a few edits in #367. Merged and released in 0.38.1.

Thanks!

derlikh-smart:
> I squashed and did a few edits in #367. Merged and released in 0.38.1.
> 
> Thanks!

Thanks for releasing it so fast!

PR #412, opened 2022-10-25
Allow selecting location to load paginations parameters from

Keats:
Hey there

For our use case we have a search with some complex filters that are hard to represent as GET parameters (we could use JSON in the query params but it's not supported either afaik so going with the simplest fix) so we use POST.
However, the PaginationMixin hardcodes the location from which to extract the params. This PR just allows the user to select another location at the decorator level.

### DIFF
--- a/docs/pagination.rst
+++ b/docs/pagination.rst
@@ -101,6 +101,9 @@ specific range of data by passing query arguments:
 
 The view function gets default values for the pagination parameters, as well as
 a maximum value for ``page_size``.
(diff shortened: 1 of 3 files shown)


lafrech:
No strong objection but couldn't you pass filters in POST and pagination parameters in query?

Keats:
That's a good point! I was just thinking of using putting all parameters in the payload but splitting them would be completely fine in our case.

PR #422, opened 2022-11-28
Supporting multiple APIs in a single app

petraszd:
Adds `Api.__init__` parameter `config_prefix`.

Solves issue https://github.com/marshmallow-code/flask-smorest/issues/252 according to [this comment](https://github.com/marshmallow-code/flask-smorest/issues/252#issuecomment-1197287924).


lafrech:
Thanks!

Looks good and quite complete. I need to carve out some time to review thoroughly as it is an important changeset, but I think this is the way to go.

Meanwhile, you may want to add a test for the error in `_get_spec_dict` to get a full coverage.

Do we really have to add that `"apis"` namespace in `app.extensions["flask-smorest"]`? Couldn't we put all versions directly, like `app.extensions["flask-smorest"] = {"": {... config for ""...}, "v1": {...config for "v1"...}`?

petraszd:
Thanks for looking at it!

> Meanwhile, you may want to add a test for the error in `_get_spec_dict` to get a full coverage.

I will do it today.


> Do we really have to add that `"apis"` namespace in `app.extensions["flask-smorest"]`? Couldn't we put all versions directly, like `app.extensions["flask-smoret"] = {"": {... config for ""...}, "v1": {...config for "v1"...}`?

We do not need that namespace. And we can easily put all versions directly into `app.extensions["flask-smoret"]`.

Previously it was `app.extensions["flask-smoret"]["ext_obj"] = CONFIG`. So I didn't want to pollute the global `app.extensions["flask-smoret"]` namespace  (for the lack of the better term) and wrapped everything with `["apis"]`.

I do not really care or think that one is better than another. So I will remove `"apis"` wrapper.

lafrech:
Hmmm... ext_obj is the Api object itself. You're right, we need a namespace for the config dicts. Could be "configs" or something rather than "apis", since we're storing configs there.

petraszd:
> Hmmm... ext_obj is the Api object itself. You're right, we need a namespace for the config dicts. Could be "configs" or something rather than "apis", since we're storing configs there.

https://github.com/marshmallow-code/flask-smorest/pull/422/files#diff-dc2102b234981f5e892b8c72b966c3a8f1e3ed65275be34b0fb274427370e08aL62

Keys are config prefixes but values are API objects (as was `ext_obj`). Are you sure about `configs`? Maybe `ext_objs`?

lafrech:
Oh, right. Sorry.

I don't mind `ext_objs` or `apis`. `apis` is correct since we store `Api` instances, but it could mean anything since everything here is about APIs. `ext_objs` is correct because we store extension object instances, but it sounds a bit cryptic.

Anyway, this is internal. Both are fine.

Or perhaps it should be

```py
{"apis": {"": {"ext_obj": ..., }, "v1": {"ext_obj": ..., }, }
```

basically the same as before with the `apis` level added to account for multiple APIs. This way we keep the possibility to add other stuff for each API.

Sorry if this was what you meant from the beginning. I've been reading a bit fast.

flask_smorest/__init__.py
```diff
@@ -92,8 +92,7 @@ def register_blueprint(self, blp, *, parameters=None, **options):
         Must be called after app is initialized.
         """
         blp_name = options.get("name", blp.name)
-        if hasattr(blp, "config_prefix"):
```
petraszd: No point in checking. Later `blp.register_views_in_doc` is called. And it is only available in smorest's blueprints.

petraszd:
@lafrech 

I've implemented it as:
```python
{"apis": {"": {"ext_obj": ..., }, "v1": {"ext_obj": ..., }, }
```

I've also added some more tests.

tests/test_api.py
```diff
@@ -1,15 +1,14 @@
 """Test Api class"""
+import apispec
```
lafrech, line 2: Why move this up there?

tests/test_api.py
```diff
@@ -1,15 +1,14 @@
...
 from flask.views import MethodView
 from werkzeug.routing import BaseConverter
-import marshmallow as ma
-import apispec
 
 from flask_smorest import Api, Blueprint
 from flask_smorest.exceptions import MissingAPIParameterError
 
-from .utils import get_schemas, get_responses
+from .utils import get_responses, get_schemas
```
lafrech, line 12: Why change the order?

flask_smorest/utils.py
```diff
@@ -137,3 +137,38 @@ def prepare_response(response, spec, content_type):
...
+    :param str config_prefix: raw prefix
+    :return: normalized prefix. Underscore is automatically added if prefix misses one.
+    """
+    result = config_prefix.strip().upper()
+    if result and not result.endswith("_"):
+        result += "_"
+    return result
+
+
+def get_config_key(*, ctx, key):
```
lafrech: Why kw only args?

flask_smorest/utils.py
```diff
@@ -137,3 +137,38 @@ def prepare_response(response, spec, content_type):
...
+
+    :param flask_smorest.Blueprint|flask_smorest.Api ctx: object
+        containing ``config_prefix`` attribute.
+    :param str key: full flask config key is constructed as ``config_prefix + key``.
+    :return: full flask config key
+    """
+    return getattr(ctx, "config_prefix", "") + key
+
+
+def get_config_value(*, app, ctx, key, default=None):
```
lafrech: Why kw only args?

lafrech (commented):
Do you think it could be possible to create the config dict at `Api` init then refer to it later?

At `Api.init_app`, get `app[config]`, extract values beginning with prefix, then store as `api.config`.

At `Blueprint` registration, keep a link to api as `self.api`, then get config from `self.api.config`, not `current_app.config`.

Pros:

- Less clumsy "get_config_value" and such code. Resulting code would be easier to read.

Cons:

- Default prefix app would have a bigger dict than other apps since all Flask config values (even those unrelated to flask-smorest) would end up in its `Api` config. Not really an issue, just a discrepancy without consequence. Well, there could be consequences in user code: if someone tries to access a Flask config parameter from the default `Api`, it will work, but then this will fail once they create a prefixed `Api`. Not a blocker, though.

- This freezes config at init. Further changes to `app.config` wouldn't be taken into account. I don't think hot modifying `app.config` was ever meant to be supported and I believed I tried not to do it in the tests. Could this be an issue?

----------------------------------------------------------------------------------------

I just realized that this feature assumes `Blueprint` instances can't be shared across `Api` instances. It wasn't obvious to me at first. Could someone want to have different APIs with common blueprints (routed to different routes, but same code)?

A way around this could be to use something in the route/request mechanism to determine, at request time, which `Api` is targetted. Then pick appropriate config. A context variable could be used to make the config available from the blueprint instance in inner view functions. I hope I'm making sense.

lafrech:
I've been thinking a little more. I think my last comment is on the right track.

Let's use a context manager to provide `current_api` (set the variable in route). This is something flask users are familiar with.

To sum it up.

- On `Api.init_app`, parse the config and extract prefixed config.
- When registering the blueprint, do something to wrap view functions to set the variable (still unclear to me, to investigate).
- When accessing config parameters, use `current_api.config`.

As I said, default prefix `Api` will be a bit special because it will contain all parameters, including Flask parameters. Users might be tempted to rely on this but other `Api` instances (prefixed ones) won't have those parameters. Unless we want to guarantee this and merge prefixed parameters with not-prefixed parameters. Prefixed ones would override non-prefixed ones. The only thing this prevents is specifying a parameter in default prefix API and not-specifying it in prefixed API (because the not-prefixed parameter would be merged and not overridden). This is only an issue if parameters have a default value, making the absence of parameter a choice. We could decide we don't set default values in the code (when reading the config) but only in the settings class in settings.py. This might be a bad pattern, though. Explicit is better than implicit, so we can expect the user to access Flask parameters in `current_app.config` and smorest ones in `current_api.config`. Besides, smorest parameters are meant for internal use so only users customizing things would be concerned. Let's keep things simple.

flask_smorest/utils.py
petraszd: I was using `get_config_value` and I myself forgot which of params is suppose to go first? `app` or `ctx`? So to force myself to use correct parameters I've made them kw only.

It is not a big deal. I can easily change it back.

---

Though, I do not know if this function is going to be useful after I will introduce a `current_api` context variable.

flask_smorest/utils.py
petraszd: Main reasoning is for `get_config_value`. There I am forcing kw so it has similar interface as `get_config_value`.

tests/test_api.py
petraszd: `A` is a first letter in alphabet. So I guess I've pressed my shortcut for `isort` :). Muscle memory. I will revert it back.

tests/test_api.py
petraszd: Same here: accidental `isort` run.

petraszd:
> Cons:
> 
> * Default prefix app would have a bigger dict than other apps since all Flask config values (even those unrelated to flask-smorest) would end up in its `Api` config.

I am strongly think that we should only extract those keys and values that are actually used in `flask-smorest`. So `current_api.config` would only contain several items.

lafrech:
> I am strongly think that we should only extract those keys and values that are actually used in `flask-smorest`. So `current_api.config` would only contain several items.

That would be nice in terms of interface. But it requires the maintenance of a centralized list of parameters (which should be extensible in case users introduce custom parameters).

flask_smorest/utils.py
lafrech: Sure. I started with a few inline comments before my longer review message.

tests/test_api.py
petraszd: OK. Turns out it is what `pre-commit` does on it's own.

petraszd:
@lafrech I've re-implemented this feature with `current_api` local proxy instead.

petraszd:
I need to look into coverage dropping a little bit. I will do that tomorrow.

petraszd:
@lafrech added new tests. Previously I've forgot to add a partial flag while checking local coverage results.

lafrech:
Just thinking out loud. Would it be a good idea to wrap the app config dict into some sort of PrefixDict proxy for each api, where the PrefixDict proxy would prefix the keys on read/write? This is closer to what you did at first, but the proxy would make it more seemless.

The benefit of this is that we'd keep the possibility to modify the config after init. And we wouldn't have to hardcode the list of parameters parsed at init. However, we'd get the downside I mentionned above, where the `""` prefix dict would get all other flask parameters.

Another approach would be to decide that all parameters start with `SMOREST_` (and `SMOREST_API_PREFIX`). This is the only reliable way to separate api parameters from other flask app parameters. Breaking change but never too late.

petraszd:
> Just thinking out loud. Would it be a good idea to wrap the app config dict into some sort of PrefixDict proxy for each api, where the PrefixDict proxy would prefix the keys on read/write? This is closer to what you did at first, but the proxy would make it more seemless.

I am looking at my changes. Actually there is no place in code base (in my PR) where anything calls `api.config` directly. Except tests. `Api.get_config_value` is used instead. IMHO:

- There is no need to store config on Api init. Api has `self._app` it can access Flask config in `get_config_value` directly.
- `PrefixDict` is a solution to a problem that I've introduced by initialising `api.config` and copying values. If there is no init config step then there is no need for a `PrefixDict`.

---

Everything is using `Api.get_config_value` anyway. That method can be a proxy. Just very simple non-magical proxy :).

---

Does it make sense? Should I just remove everything else besides `get_config_value` and `current_api`?

lafrech:
My point (my intuition) was that `api.config` could be the proxy I describe above. It would look nicer than `api.get_config_value`. `api.config` could also be a property generating the dict on-the-fly (beware of perf issues). The point is you could write `api.config[SOME_PARAMETER]`.

petraszd:
Usual usage of `config` is following:

https://github.com/marshmallow-code/flask-smorest/blob/master/flask_smorest/spec/__init__.py#L182

```python
title = self._app.config.get("API_TITLE", title)  # `title` is a function argument.
```

So just using `api.config[SOME_PARAMETER]` is not helpful. Even one-liner `api.config["API_TITLE"] or title` would not work (What if I explicitly set `API_TITLE` as an empty string).

Usual usage of `api.config` would be `api.config.get(key, default_value)`. And I think that `api.get_config_value` is the same from a beauty perspective.

I can implement `PrefixDict`. It is probably not that hard (I will regret claiming it 😀). But I think it is an unnecessary magic.

lafrech:
> Usual usage of api.config would be `api.config.get(key, default_value)`.

Good. The point is that config is expected to be is a dict. This is what I propose. Be it a proxy or the property trick described above.

I realize it does not make a huge difference from `get_config_value` and it is pointless to argue about beauty. I just figured from the caller perspective it would make a nicer API because by analogy with flask.config, people would expect a config dict without having to "learn" a new function.

To be honest, I haven't been giving this enough thought. Sorry if I make you go back and forth changing my mind along the way.

petraszd:
OK. Got it. I will make it a dict (I new I should not have said that it will be easy to implement. Now I've jinxed it).

I was thinking more about internal usage of `api.config` and you were thinking about external usage. From external usage it being a dict like makes more sense.

flask_smorest/prefixed_config.py
```diff
@@ -0,0 +1,16 @@
+from collections.abc import Mapping
+
+
+class PrefixedConfigProxy(Mapping):
```
petraszd: Maybe I should move it to `utils.py`?

petraszd:
@lafrech I've added prefixed config proxy.

flask_smorest/prefixed_config.py
lafrech: Yes. And you cold make it more generic, like `PrefixedMappingProxy`, taking proxied `dict` as input rather than `app`.

Then you'd do

```py
self.config = PrefixedMappingProxy(app.config, self.config_prefix)
```

petraszd: Done.

petraszd:
@lafrech : any news on this one?

lafrech:
Sorry. Just needs time. Haven't had much personal time lately.

flask_smorest/utils.py
```diff
@@ -137,3 +137,38 @@ def prepare_response(response, spec, content_type):
...
+    :return: normalized prefix. Underscore is automatically added if prefix misses one.
+    """
+    result = config_prefix.strip().upper()
+    if result and not result.endswith("_"):
+        result += "_"
+    return result
+
+
+class PrefixedMappingProxy(abc.Mapping):
+    """Mapping to proxy an another mapping using a prefix
```
lafrech: s/an another/another/

flask_smorest/utils.py
```diff
@@ -137,3 +137,38 @@ def prepare_response(response, spec, content_type):
...
+
+    .. code-block:: python
+        some_dict = {"foobar_key1": 1, "foobar_key2": 2}
+        assert some_dict["key1"] == 1
+        assert some_dict["key2"] == 2
```
lafrech, line 162: I don't think this code sample really helps. I find it confusing.

An explanation sentence could be fine. Or a code sample with the class instantiation.

flask_smorest/etag.py
```diff
@@ -142,14 +143,26 @@ def check_etag(self, etag_data, etag_schema=None):
...
 
+    def _is_etag_enabled(self, api=None):
+        """Return True if ETag feature enabled api-wise"""
+        api = api or current_api
+        # Do not check `api is None`. `current_api` is a LocalProxy. It is never `None`.
+        if not api:
+            raise CurrentApiNotAvailableError(
+                "Can't access current API config. "
+                "Make sure that request context is available "
+                "and/or ETag is checked inside flask_smorest views."
```
lafrech: I don't understand that last line.

flask_smorest/globals.py
```diff
@@ -0,0 +1,20 @@
+from flask import current_app, has_request_context, request
+from werkzeug.local import LocalProxy
+
+
+def _find_current_api():
+    if has_request_context() and "flask-smorest" in current_app.extensions:
+        apis = current_app.extensions["flask-smorest"]["apis"].values()
```
lafrech: Could we just call current_app and let the Flask exception bubble up, rather than implement our own exception?

Do we need to manage the case where "flask-smorest" is not in app.extensions?

flask_smorest/globals.py
```diff
@@ -0,0 +1,20 @@
...
+        blueprints = set(request.blueprints)
```
lafrech: `request.blueprints` requires werkzeug 2.0.1. setup.py needs to be updated.

flask_smorest/globals.py
```diff
@@ -0,0 +1,20 @@
...
+        for api in (x["ext_obj"] for x in apis):
+            if api._registered_blueprint_names & blueprints:
+                return api
+    return None
```
lafrech, line 13: Iterating and matching on names. This makes me cringe a bit.

I expected this part to be simpler but it was just an intuition. I haven't been digging into this myself, so maybe that's how it is.

Could blueprint name clash occur?

Would it make sense to set a {blp_name: api_obj} mapping in app.extensions["flask-smorest"] when registering blueprints so that the lookup here can be replaced with a simple key access? Or does it sound too hackish?

lafrech (commented):
This looks good.

It is an important change. If I was doing it, I'd abuse `git rebase -i` to rewrite it into two or three meaningful commits, for ease of review and nicer history. If it is too much to ask, I might do it myself at some point before merging.

There are still details I'd like to review, but here's a bunch of comments already.

Thanks.

flask_smorest/utils.py
petraszd: Oh. I actually forgot to use `PrefixedMappingProxy` and just used a simple dict :). Fixed. I think now it is clearer.

flask_smorest/etag.py
petraszd: Removed. I think request context is good enough.

flask_smorest/globals.py
petraszd: Probably not. And if `KeyError` would complain about `"flask-smorest"` it is pretty clear indicator that you need a "flask-smorest".

petraszd: Removed that check.

flask_smorest/globals.py
petraszd: > Would it make sense to set a {blp_name: api_obj} mapping in app.extensions["flask-smorest"] when registering blueprints so that the lookup here can be replaced with a simple key access? Or does it sound too hackish?

I think it is better than a set and for-eaching through all apis. I've updated PR to use `{blp_name: api_obj}` instead. Thanks.

> Could blueprint name clash occur?

No. It raises ValueError.

https://github.com/pallets/flask/blob/main/src/flask/blueprints.py#L383-L391
```python
        if name in app.blueprints:
            ...
            raise ValueError(f"The name '{self_name}' is already registered...")
```

> I expected this part to be simpler but it was just an intuition. I haven't been digging into this myself, so maybe that's how it is.

It is annoying because a blueprint can be registered with multiple APIs. Sometimes even indirectly. For example: you can register a blueprint with one API and then nest it into other blueprint. Which in turn would be registered with some another API.

So I can not mess with a passed blueprint. Which means that I would need to add some kind of `BlueprintProxy` to intercept requests, set `current_api`, and pass them to an original blueprint. Would work I guess. But it would mean that each API request would have an additional function call. And I do not think that `current_api` would be needed 99% of time. So I would rather make getting `current_api` slightly slower than to add an additional frame stack to each API request.

petraszd:
> It is an important change. If I was doing it, I'd abuse git rebase -i to rewrite it into two or three meaningful commits, for ease of review and nicer history. If it is too much to ask, I might do it myself at some point before merging.

@lafrech let's get to the point were you would feel that changes are good to be merged into the master branch. And then I can create a new PR with a single commit.

lafrech:
Hi. Sorry again about the delay. I rebased and made a few changes: #485.

Would you like to review?

lafrech:
Superseded by #485.

Thanks for the work on this.

Issue #238, opened 2021-04-08
Using a dictionary as a Schema


aprilahijriyan:
i get error when using dictionary as Schema.

The dictionary that I use:

```python
@use_kwargs({"limit": fields.Integer(), "offset": fields.Integer()}, location="query")
```

the following is the error log.

```
Traceback (most recent call last):
  File "/home/Projects/zemfrog/venv/bin/flask", line 8, in <module>
    sys.exit(main())
  File "/home/Projects/zemfrog/venv/lib/python3.6/site-packages/flask/cli.py", line 967, in main
    cli.main(args=sys.argv[1:], prog_name="python -m flask" if as_module else None)
  File "/home/Projects/zemfrog/venv/lib/python3.6/site-packages/flask/cli.py", line 586, in main
    return super(FlaskGroup, self).main(*args, **kwargs)
  File "/home/Projects/zemfrog/venv/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/Projects/zemfrog/venv/lib/python3.6/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/Projects/zemfrog/venv/lib/python3.6/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/Projects/zemfrog/venv/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/Projects/zemfrog/venv/lib/python3.6/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/Projects/zemfrog/venv/lib/python3.6/site-packages/flask/cli.py", line 425, in decorator
    with __ctx.ensure_object(ScriptInfo).load_app().app_context():
  File "/home/Projects/zemfrog/venv/lib/python3.6/site-packages/flask/cli.py", line 392, in load_app
    app = locate_app(self, import_name, None, raise_if_not_found=False)
  File "/home/Projects/zemfrog/venv/lib/python3.6/site-packages/flask/cli.py", line 240, in locate_app
    __import__(module_name)
  File "/home/Projects/zemfrog/demo/app/wsgi.py", line 3, in <module>
    app = create_app("wsgi")
  File "/home/Projects/zemfrog/zemfrog/app.py", line 79, in create_app
[...]

PR #423, opened 2022-11-28
support dict as Schema

### LINKED ISSUE #180: Using a dict of fields for arguments
In the docs, I see there is a way to do: `@blp.arguments(PetSchema)`
Is there a way to simply do: `@blp.arguments({'name': ma.fields.Str()})` ?

I was trying to port some of my projects from flaks-apispec to flask-smorest and in flask-apispec I was doing: `@use_kwargs({'name': ma.fields.Str()})` a lot as that seemed generic and easy to define for my endpoints with simple query params.
I didn't see a similar thing in smorest, and was wondering if I have to define a schema every time ?

I don't like schemas too much because they become a bit verbose for 1 or 2 arguments ... and then in cases where I need borh a query-param and a body as arguments (POST with query params) I need to define 2 schemas - which seems to be confusing

OmmyZhang:
Sometimes it's not necessary to define a schema class for each API.

This PR support `@blp.arguments({'arg': ma.fields.Str()})` and `@blp.response(200, {'arg': ma.fields.Str()})`

close #180  #238

### DIFF
--- a/flask_smorest/arguments.py
+++ b/flask_smorest/arguments.py
@@ -4,6 +4,7 @@ from copy import deepcopy
 from functools import wraps
 import http
 
+import marshmallow as ma
 from webargs.flaskparser import FlaskParser
 
 from .utils import deepupdate
@@ -28,8 +29,8 @@ class ArgumentsMixin:
     ):
         """Decorator specifying the schema used to deserialize parameters
 
-        :param type|Schema schema: Marshmallow ``Schema`` class or instance
-            used to deserialize and validate the argument.
+        :param type|Schema|dict schema: Marshmallow ``Schema`` class or instance
+            or dict used to deserialize and validate the argument.
         :param str location: Location of the argument.
         :param str content_type: Content type of the argument.
             Should only be used in conjunction with ``json``, ``form`` or
@@ -56,6 +57,8 @@ class ArgumentsMixin:
 
         See :doc:`Arguments <arguments>`.
         """
+        if isinstance(schema, dict):
+            schema = ma.Schema.from_dict(schema)
         # At this stage, put schema instance in doc dictionary. Il will be
         # replaced later on by $ref or json.
         parameters = {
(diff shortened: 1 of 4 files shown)


lafrech:
Thanks. This PR proves it is feasible with little modifications.

I'm not so happy with the test in `resolve_schema_instance`, but I don't have any obvious alternative.

Also it looks from your comment like this function accepts dictionaries already, while it was not explicitly intended to. Perhaps something I overlooked that worked by luck.

I shall look into this and perhaps rework this to make things explicit.

It's a pity we do the `from_dict` call in arguments while this is taken care of by webargs already, but we need it for the docs and getting it back from webargs would be tricky, if even possible. Not an issue.

OmmyZhang:
> Also it looks from your comment like this function accepts dictionaries already, while it was not explicitly intended to.

  Yes, I find this in https://github.com/marshmallow-code/flask-smorest/commit/f5b7212602203208187671c59e7442efd7cec327 when I try to figure out why `ResponseMixin` says `:param schema schema|str|dict:`.

Personally, I suggest changing this usages. It's so confusing.

lafrech:
Yes. There's something a bit fishy in this. Perhaps I should review it, especially now that `alt_response` is available.

I'd rather sort this out before going forward with this PR trying to find workarounds for a design issue that ought to be addressed first.

lafrech:
I looked into this tonight. Passing the schema doc as `dict`, which is a pretty uncommon use case, should probably be achieved using another parameter, like `schema_doc` or something.

Then any `dict` passed as `schema` would be expected to be a `dict` of fields to feed `from_dict`.

lafrech:
Back to this. Since passing doc as dict is neither really documented (apart from the docstring type) nor tested, I'm open to dropping the case. If anyone complains we can always figure out a way.

I've been trying to implement this on my own taking inspiration in this PR and I realized that generated schemas have a default name which yields warnings in apispec. See https://github.com/marshmallow-code/flask-smorest/issues/180#issuecomment-1681115460. Let's discuss this issue there.

Issue #71, opened 2019-05-09
[RFC] Deprecate Blueprint.doc decorator


lafrech:
The `Blueprint.doc` decorator is meant to add documentation that can't be inferred from the code or added by dedicated decorators. It allows the user to pass manual documentation that will be deep-merged with the automatically generated documentation. Unfortunately, it is not the silver bullet one would expect. It suffers from a few intrinsic issues.

We could try to play smart and improve the way the structure received by `doc` is used, but I'd rather add dedicated decorators or other explicit means to allow the user to pass documentation in a structured fashion, with a clear and thin abstraction layer.

I'd like to use this issue to list known issues about `Blueprint.doc` and use cases that should be covered by other means. The goal being to make it useless, and to deprecate it if possible.

### Known issues

- It can't really document arguments as deep-merging lists doesn't work so it conflicts with auto-generated documentation.
- When adding extra documentation for a response, the user must use the same type for the status code than the one used in the `response` decorator (https://github.com/marshmallow-code/flask-smorest/issues/60#issuecomment-487856593).
- It doesn't handle the OASv2 / v3 transparently, while the other decorators are pretty good at it, allowing the user to change the OAS version with a configuration parameter using the same codebase. For instance, it doesn't add `['content']['application/json']` when using v3 which makes it more verbose to the user.

### Use cases

- [x] Add summary/description: Those can be passed in the view function docstring.
- [x] Add examples: Can now be achieved with `response`/`arguments` decorators.
- [ ] Document specific API features (e.g. security information). Since this is quite specific, I think it could be done by a custom decorator defined in user code. Typically, when using a decorator to manage accesses to resources (generally from a flask third-party library), this decorator should be wrapped in a custom decorator that adds the information to the docs (see https://github.com/marshmallow-code/flask-smorest/issues/36#issuecomment-460826257 for an example). This means the `_apidoc` attribute mechanism becomes public API.
- [x] Document / add extra-information to parameters. The only real need was for path parameters. Documentation for path parameters can be passed in `Blueprint.route`. (See #23)
- [x] Document / add extra-information to responses. We could add a specific decorator for that. If we assume a view function can only have one normal response and other responses are errors (or error-likes such as redirections or "not modified"), then that decorator could be called "error". (See #44)

dougthor42:
It's likely that most of these issues would be solved by #49, no? Users could simply set `docstring_sep=None` on the blueprint and then whatever they add to the function docstring will be included in `description`.

[...]

PR #436, opened 2022-12-31
Add plugin system for Blueprints and Api

Aesonus:
Introduces a plugin system to add functionality using composition instead of inheritance. This requires the plugin to be registered on the blueprint. See test_plugin.py for more details.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -25,3 +25,4 @@ Contributors (chronological)
 - Choudhury Noor `@Cnoor0171 <https://github.com/Cnoor0171>`_
 - Dmitry Erlikh `@derlikh-smart <https://github.com/derlikh-smart>`_
 - 0x78f1935 `@0x78f1935 <https://github.com/0x78f1935>`_
+- Cory Laughlin `@Aesonus <https://github.com/Aesonus>`_
--- a/flask_smorest/__init__.py
+++ b/flask_smorest/__init__.py
@@ -88,6 +88,9 @@ class Api(APISpecMixin, ErrorHandlerMixin):
 
         self._app.register_blueprint(blp, **options)
 
+        for bp_plugin in getattr(blp, "_smore_plugins", []):
+            bp_plugin.visit_api(self)
+
         # Register views in API documentation for this resource
         blp.register_views_in_doc(
             self,
--- a/flask_smorest/blueprint.py
+++ b/flask_smorest/blueprint.py
@@ -72,6 +72,9 @@ class Blueprint(
 
         self.description = kwargs.pop("description", "")
 
+        # This is where smore plugins are stored
+        self._smore_plugins = kwargs.pop("smore_plugins", [])
+
         super().__init__(*args, **kwargs)
 
         # _docs stores information used at init time to produce documentation.
@@ -97,6 +100,7 @@ class Blueprint(
             self._prepare_response_doc,
             self._prepare_pagination_doc,
             self._prepare_etag_doc,
+            *[plugin.register_method_docs for plugin in self._smore_plugins],
         ]
 
     def add_url_rule(
--- a/flask_smorest/plugin/abc.py
+++ b/flask_smorest/plugin/abc.py
@@ -1,0 +1,27 @@
+import abc
+
+
+class Plugin(abc.ABC):
+    """Abstract base class to structure smore plugins"""
+
+    @abc.abstractmethod
+    def register_method_docs(self, doc, doc_info, *, api, spec, **kwargs):
+        """
+        Call when the views are registered in doc
+
+        :param  dict doc: The current operation doc
+        :param dict doc_info: Doc info stored by decorators
+        :param Api api: The Api() instance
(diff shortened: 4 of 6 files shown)


Aesonus:
Hello,
I found myself reading through the discussion over in issue #71 and thought I might take a crack at making a simpler way for users to add their own decorators and apidoc without having to rely on Blueprint.doc. This PR aims to get the ball rolling in that direction by moving away from adding more mixin types on Blueprint and moving towards a compositional pattern. Granted, this PR does still rely on public access to `func._apidoc` on decorator callable to add API spec, but that aside, this would be a start.

I also cast a vote for calling plugins for this purpose "Smore Plugins".

lafrech:
Thanks for the contribution and sorry for not answering any sooner.

Please allow me a little time to give this the attention it deserves. (Don't worry about test coverage at this stage.)

Aesonus:
Not a problem at all. Thanks.

lafrech:
I think I get the point. But it's not clear to me what it allows that can't be done without it. Admittedly, documenting a security decorator is something users have been wondering how to do and I have never really taken the time to lay this down. But I'm not sure this plugin interface is required and how much it helps.

You may be interested in what I do in [bemserver-api](https://github.com/BEMServer/bemserver-api). This is the way I recommend over using `blp.doc`.

In `extensions` package, there's a `smorest` module, in which I subclass `Api` to document the security scheme and I subclass `Blueprint` to add the `login_required` method and add a doc callback. I agree this is not really obvious, it deserves a better interface and a plugin interface such as this one seems to be a step in the right direction. But it still exposes non-obvious stuff. For instance, `register_method_docs`'s signature may require users to understand the internals to an extend that they could do what I'm doing in bemserver-api without the plugin interface.

I realize this answer might not be satisfying but unless we manage to produce a very simple interface (is this even feasible?) I'm tempted to cowardly leave things as is (and point advanced users to real-life examples).

I understand your point about compositing vs. subclassing. But there are so many advanced features that need subclassing that I always recommend to subclass when starting a new app, just to make it easier on the long run. And I find it nicer in the resource pages to have all features as `blp` methods (totally opinionated, I admit).

This is not a definitive no. Just thinking out loud that this might not be needed (and I'd rather not add uneeded API surface).

Aesonus:
Thank you for taking the time to look this over and give your thoughts.

> I think I get the point. But it's not clear to me what it allows that can't be done without it.

In all honesty, nothing at this point. I did not really intend for this to add functionality but rather to reduce the amount of knowledge a user may need about `Blueprint` to extend functionality using subclassing. While I was writing this PR, I had to take a good look at the source code to figure out how all the docs were generated, leading me to the `Blueprint._prepare_doc_cbks` attribute. This would require users wanting to extend the functionality knowing about the internal workings of `Blueprint` and `Api`. That brings me to your other point.

> In extensions package, there's a smorest module, in which I subclass `Api` to document the security scheme and I subclass `Blueprint` to add the `login_required` method and add a doc callback. I agree this is not really obvious, it deserves a better interface and a plugin interface such as this one seems to be a step in the right direction. But it still exposes non-obvious stuff. For instance, `register_method_docs`'s signature may require users to understand the internals to an extend that they could do what I'm doing in bemserver-api without the plugin interface.

I can agree somewhat, however, what you do in bemserver-api would require the user to understand that `Blueprint` relies on `Blueprint._prepare_doc_cbks` to add the additional api from `_apidoc` to the spec. They would need to know to append the `_prepare_auth_doc` method to `Blueprint._prepare_doc_cbks` and also to extend the `Api` class as well to document the security schema. What I have submitted here eliminates those extra steps. While I agree that this plugin system does the same thing bemserver-api does, I am of the opinion that what I have done reduces the number of steps to add additional functionality, and is particularly advantageous when adding more than one extension.

For instance, if I was wanting to have security schema and also give each endpoint an `operationId` using decorators, I would end up adding more methods to the same extended `Blueprint` class, violating the single responsibility principle. By separating each additional extension to its own class, it avoids that problem. That being said, it does take the decorators out of the `Blueprint` class. That could be an easy fix using `__getattr__` in the `Blueprint` class to proxy to the plugin decorator method, and registering plugins on a class attribute rather than an instance attribute, such that every `Blueprint` instance can use the registered plugins. I may revise this pull request to reflect that.

> I realize this answer might not be satisfying but unless we manage to produce a very simple interface (is this even feasible?) I'm tempted to cowardly leave things as is (and point advanced users to real-life examples).

I think it is feasible, though it is not clear to me how at this point. I'll have to get back to you on that.

In any case, I appreciate your response. It certainly reveals quite a bit about the paradigms used. I will see what else I can come up with. I really enjoy this project (I use it a lot), so I am happy to help make it better.

lafrech:
Thank you for the nice feedback. And for the constructive help. This is definitely not a no go and I'd be happy if we could go further and actually provide a nice interface for the reasons you explained above.

I don't have enough time to put into this but you seem to have a good vision about it now and I'll try to be as reactive as can be.

PR #485, opened 2023-04-04
Support multiple APIs in a single app

lafrech:
Rebase / rework of @petraszd's #422.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -25,3 +25,4 @@ Contributors (chronological)
 - Choudhury Noor `@Cnoor0171 <https://github.com/Cnoor0171>`_
 - Dmitry Erlikh `@derlikh-smart <https://github.com/derlikh-smart>`_
 - 0x78f1935 `@0x78f1935 <https://github.com/0x78f1935>`_
+- One Codex, Inc. `@onecodex <https://github.com/onecodex>`_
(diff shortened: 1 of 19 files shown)


lafrech:
I'm still not convinced there couldn't be name clashes in `_find_current_api` if someone created a blueprint with a name containing a dot that would match a nested blueprint name. Admittedly, this is a bit unlikely, but...

petraszd (approved):

petraszd:
@lafrech 

It is impossible to create a blueprint with a dot `.` in it's name:
https://github.com/pallets/flask/blob/2.0.x/src/flask/blueprints.py#L198-L199

lafrech:
Oh, good. I thought I had tested this. I guess I hadn't.

Thanks for reviewing.

lafrech:
I just performed a last review and a few minor edits. This is good to go.

Before merging, I'd like to tests against my code base at work next week, just in case, although I'm sure it's all non-breaking.

I'll hopefully release after that.

lafrech:
It is worth noting that the `Api` instance is now accessible at the following path:

```py
app.extensions["flask-smorest"]["apis"][""]["ext_obj"]
```

Users accessing the `Api` object at `app.extensions["flask-smorest"]["ext_obj"]` will have to adapt their code.

We could add a deprecated shortcut there, but frankly I don't think it is worth it.

PR #542, opened 2023-08-17
#541 Fix order of automatic path parameters extracted from Flask route to be consistently left to right

### LINKED ISSUE #541: Inconsistent order of parameters when using rule.arguments from werkzeug inside FlaskPlugin
I'm using this with flask-smorest. In https://flask-smorest.readthedocs.io/en/latest/openapi.html#enforce-order-in-openapi-specification-file, you say to use "ordered = True" so they're ordered and I can do snapshot-testing on the openapi json output.

~~But in this line:https://github.com/marshmallow-code/apispec/blob/dev/src/apispec/ext/marshmallow/common.py#L54, `schema._declared_fields` is not an ordered dict, but a simple dict. The `schema.fields` is ordered though.~~

~~Do you have any idea how to fix this even with classes instead of just objects?~~

See https://github.com/marshmallow-code/flask-smorest/issues/541 for correct issue

Regards,
Dorian

ddorian:
Fixes #541

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -26,3 +26,4 @@ Contributors (chronological)
 - Dmitry Erlikh `@derlikh-smart <https://github.com/derlikh-smart>`_
 - 0x78f1935 `@0x78f1935 <https://github.com/0x78f1935>`_
 - One Codex, Inc. `@onecodex <https://github.com/onecodex>`_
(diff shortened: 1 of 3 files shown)


lafrech:
Thanks!

Released in 0.42.1.

PR #549, opened 2023-09-05
Biginteger field

ddorian:
I was using https://github.com/schemathesis/schemathesis to test my openapi, and the first thing it did was to send a out-of-range bigint, that was accepted by python, but actually errored in PostgreSQL (`sqlalchemy.exc.DataError: (psycopg.errors.NumericValueOutOfRange) bigint out of range`).

So I created a BigInteger field that checks that.

### DIFF
--- a/flask_smorest/fields.py
+++ b/flask_smorest/fields.py
@@ -13,3 +13,18 @@ class Upload(ma.fields.Field):
     def __init__(self, format="binary", **kwargs):
         self.format = format
         super().__init__(**kwargs)
+
+
+class BigInteger(ma.fields.Integer):
+    """A bigint field."""
+
+    def _validated(self, value):
+        """Validate that the number is inside 64bit bigint."""
+        new_value = super()._validated(value)
+        invalid_bigint = (new_value > (2**63 - 1)) or (new_value < (-(2**63) + 1))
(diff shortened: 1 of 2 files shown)


lafrech:
Thanks for bringing this up and sharing your implementation.

I won't integrate this here. It would be better suited for marshmallow itself, it is not specific to flask-smorest.

Regarding the implementation itself, couldn't you use a `Range` validator?

I'm wondering if it makes sense to integrate a PostgreSQL specific field in marshmallow if it is as easy for the user as specifying an existing validator.

Perhaps marshmallow-sqlalchemy could add that validator automatically or integers.

ddorian:
> Regarding the implementation itself, couldn't you use a Range validator?

Is there a way to add a "default" validator to all instances of a class? Otherwise I'd have to repeat it often, no?

> Perhaps marshmallow-sqlalchemy could add that validator automatically or integers.

Maybe.

Thank you.

lafrech:
You may subclass the field, like you already do, to add this validator by default.

lafrech:
My point was that it seemed better to me to use the existing validator that write comparisons.

Regarding marshmallow integration, see See https://github.com/marshmallow-code/marshmallow/issues/2182.

PR #572, opened 2023-10-25
allow flask 3.0.x and werkzeug 3.0.x

### LINKED ISSUE #573: Werkzeug Security Issue
Based on Nexus IQ Server there is a critical issue with Werkzeug version 2.3.7 and the version 3.0.1 has to be installed which your package doesn't allow.

Security level 8 -> CVE-2023-46136

sjhewitt:
i had a look through the code and didn't see anything affected by any of the breaking changes

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,14 +1,14 @@
 #!/usr/bin/env python3
 
-from setuptools import setup, find_packages
+from setuptools import find_packages, setup
 
 EXTRAS_REQUIRE = {
     "tests": [
         "pytest==7.4.2",
         "pytest-cov==4.1.0",
         "coverage==7.3.2",
-        "werkzeug==2.3.7",
-        "flask==2.3.3",
+        "werkzeug==3.0.1",
+        "flask==3.0.0",
         "marshmallow==3.20.1",
         "webargs==8.3.0",
         "apispec==6.3.0",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -1,14 +1,14 @@
 #!/usr/bin/env python3
 
-from setuptools import setup, find_packages
+from setuptools import find_packages, setup
```
lafrech, line 3: I don't mind, but out of curiosity, is this automatic reordering from a tool like PyCharm or something?

setup.py
```diff
@@ -66,8 +66,8 @@
     },
     python_requires=">=3.8",
     install_requires=[
-        "werkzeug>=2.0.1,<3",
-        "flask>=2.0.1,<3",
+        "werkzeug>=2.0.1,<3.1",
+        "flask>=2.0.1,<3.1",
```
lafrech, line 70: I'd set the max to 4.0. Things shouldn't break in between.

lafrech (changes requested):
Thanks for your help on this.

I feel like we ought to add a specific tox env to test minimum versions of dependencies, like we did on webargs for instance, to be sure we don't use new Flask features without realizing it because the tests only test the latest version.

https://github.com/marshmallow-code/webargs/blob/dev/tox.ini

If you'd like to tackle this, you're more than welcome. Otherwise, let's just set 4.0 as max as per my comment and merge this for now. The minimum dependency versions test issue shouldn't block this.

lafrech:
Superseded by #576.

Thanks anyway.

Issue #578, opened 2023-11-14
Missing response header in OpenAPI documentation


drcpu-github:
When generating OpenAPI documentation, additional headers in the response decorator are overwritten whenever the paginate decorator is used. This appears to be caused by the `_document_pagination_metadata` in `pagination.py` which overwrites the `resp_doc["headers"]` structure.

A simple fix could be something like this:
```py
def _document_pagination_metadata(self, spec, resp_doc):
    """Document pagination metadata header

    Override this to document custom pagination metadata
    """
    if "headers" not in resp_doc:
        resp_doc["headers"] = {}
    resp_doc["headers"].update({
        self.PAGINATION_HEADER_NAME: "PAGINATION"
        if spec.openapi_version.major >= 3
        else PAGINATION_HEADER
    })
```

lafrech:
Yes, I guess you're right.

Would you like to submit a PR with this fix and a test?

drcpu-github:
Sure thing, I created one.

lafrech:
Fixed in #579.

PR #579, opened 2023-11-16
Fix OpenAPI docs generation to include missing response headers when pagination is used

drcpu-github:
PR for issue #578.

Added the code I showed in the issue and a simple test to show that it works. Not sure if you expect more from a test than this, but let me know.

### DIFF
--- a/flask_smorest/pagination.py
+++ b/flask_smorest/pagination.py
@@ -276,11 +276,13 @@ class PaginationMixin:
 
         Override this to document custom pagination metadata
         """
-        resp_doc["headers"] = {
-            self.PAGINATION_HEADER_NAME: "PAGINATION"
-            if spec.openapi_version.major >= 3
-            else PAGINATION_HEADER
-        }
+        resp_doc.setdefault("headers", {}).update(
+            {
+                self.PAGINATION_HEADER_NAME: "PAGINATION"
+                if spec.openapi_version.major >= 3
+                else PAGINATION_HEADER
+            }
+        )
 
     def _prepare_pagination_doc(self, doc, doc_info, *, spec, **kwargs):
         operation = doc_info.get("pagination")
--- a/tests/test_pagination.py
+++ b/tests/test_pagination.py
@@ -479,6 +479,39 @@ class TestPagination:
             assert parameters[3]["schema"]["minimum"] == 1
             assert parameters[3]["schema"]["maximum"] == 100
 
+    # Non-regression test for https://github.com/marshmallow-code/flask-smorest/issues/578 # noqa: E501
+    @pytest.mark.parametrize("openapi_version", ("2.0", "3.0.2"))
+    def test_pagination_doc_preserves_other_headers_doc(self, app, openapi_version):
+        app.config["OPENAPI_VERSION"] = openapi_version
+        api = Api(app)
+        blp = Blueprint("test", __name__, url_prefix="/test")
+
+        @blp.route("/")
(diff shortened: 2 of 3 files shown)


flask_smorest/pagination.py
```diff
@@ -276,11 +276,15 @@ def _document_pagination_metadata(self, spec, resp_doc):
...
         Override this to document custom pagination metadata
         """
-        resp_doc["headers"] = {
-            self.PAGINATION_HEADER_NAME: "PAGINATION"
-            if spec.openapi_version.major >= 3
-            else PAGINATION_HEADER
-        }
+        if "headers" not in resp_doc:
+            resp_doc["headers"] = {}
+        resp_doc["headers"].update(
```
lafrech: I prefer the setdefault form. That's the form I use everywhere in the code.

```py
resp_doc.setdefault("headers", {}).update(...)
```

tests/test_spec.py
```diff
@@ -178,6 +178,29 @@ def test_get(val):
             "schema": {"$ref": "#/components/schemas/PaginationMetadata"},
         }
 
+    def test_all_headers_are_documented(self, app):
```
lafrech: I'm tempted to add a comment such as 

```py
# Non-regression test for https://github.com/marshmallow-code/flask-smorest/issues/578
```

to make the test intent more explicit.

lafrech: We could put this in pagination tests and make the name reflect the fact that we're checking other headers are not dropped as a non-reg test for this specific issue.

I'd be happy to enlarge the scope to check all sorts of combination and leave the test here. But this is not what we do here and I'm not sure it's worth trying.

For now, I'd move this to `test_pagination.py` and call it something like `test_pagination_doc_preservers_other_headers_doc`.

lafrech (changes requested):
Thanks. That's what I was expecting. 

Once minor comments are addressed, this is good to go.

tests/test_spec.py
drcpu-github: I did add a `#noqa: E501` to stop flake8 complaining about the length of that line.

drcpu-github: Funilly enough, I first had that test in `test_pagination.py`, but then moved it because I felt it better fit in `test_spec.py` since it was about documentation. In any case, I moved and renamed it.

drcpu-github:
I have applied all suggestions, but did not yet squash the branch so you can check it more easily. I'll do that once you are happy with the changes.

tests/test_pagination.py
```diff
@@ -479,6 +479,39 @@ def func():
             assert parameters[3]["schema"]["minimum"] == 1
             assert parameters[3]["schema"]["maximum"] == 100
 
+    # Non-regression test for https://github.com/marshmallow-code/flask-smorest/issues/578 # noqa: E501
+    @pytest.mark.parametrize("openapi_version", ("2.0", "3.0.2"))
+    def test_pagination_doc_preservers_other_headers_doc(self, app, openapi_version):
```
lafrech: s/preservers/preserves/

lafrech (changes requested):
Just a typo to fix and you can squash.

Thanks!

drcpu-github:
> Just a typo to fix and you can squash.
> 
> Thanks!

Done!

lafrech (approved):

lafrech:
Released in 0.42.3. Thanks!

PR #591, opened 2023-12-29
Support python 3.12

TGoddessana:
### Support Python 3.12

- added tox test environment 3.12
- changed github workflows run test with python3.8 to python3.12
- added classifires in `setup.py`

Since all the test code passed on python3.12, I just modified the CI to test python3.12.


lafrech:
Thanks.

Note that apispec and webargs are not tested on 3.12 yet. We need to make the same changes there.

PR #593, opened 2024-01-06
Fix type-hint in paginate method

jtait:
The type-hint in the current docs require an instance, not a class.  This PR corrects that.


lafrech:
Right. We should implement actual type hints + mypy tests. Anyone willing to help with this is welcome.

Thanks for the correction.

PR #635, opened 2024-04-18
Update arguments.rst

juur:
Declare support for json_or_form, which is referenced further down but not in the locations list.


lafrech:
Thanks!

PR #644, opened 2024-05-04
Improve errorhandler

TGoddessana:
## There are a few issues... (shouldn't be merged yet)

1. The `abort(code)` works, but raise ApiException() doesn't work with the json response, in my opinion, raising an `ApiException` should automatically result in a 500 json handling, but it's not.
The `flask_abort()` and `api_abort()` use cases work anyway... I just want to handle them elegantly. (What should be the behavior if the user raises `ApiException()` instead of `flask_smorest.abort()`?)

2. I looked at the method you mentioned in the original issue(re-define all werkzeug exceptions), and my guess is that since we are inheriting and defining some exceptions from Flask-Smorest, it would be better to explicitly inherit them all, even if it makes the code a bit longer (I'd love to hear your opinion).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -157,4 +157,4 @@ cython_debug/
 #  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
-#.idea/
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -157,4 +157,4 @@ cython_debug/
 #  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
-.idea/
+#.idea/
--- a/src/flask_smorest/exceptions.py
+++ b/src/flask_smorest/exceptions.py
@@ -7,11 +7,151 @@ class FlaskSmorestError(Exception):
     """Generic flask-smorest exception"""
 
 
+# Non-API exceptions
 class MissingAPIParameterError(FlaskSmorestError):
     """Missing API parameter"""
 
 
-class NotModified(wexc.HTTPException, FlaskSmorestError):
+class CurrentApiNotAvailableError(FlaskSmorestError):
+    """`current_api` not available"""
+
+
+# API exceptions
+class ApiException(wexc.HTTPException, FlaskSmorestError):
+    """A generic API error."""
+
+    @classmethod
+    def get_exception_by_code(cls, code):
+        for exception in cls.__subclasses__():
+            if exception.code == code:
+                return exception
+        return InternalServerError
+
+
+class BadRequest(wexc.BadRequest, ApiException):
+    """Bad request"""
+
+
+class Unauthorized(wexc.Unauthorized, ApiException):
+    """Unauthorized access"""
+
+
+class Forbidden(wexc.Forbidden, ApiException):
+    """Forbidden access"""
+
+
+class NotFound(wexc.NotFound, ApiException):
+    """Resource not found"""
+
+
+class MethodNotAllowed(wexc.MethodNotAllowed, ApiException):
+    """Method not allowed"""
+
+
+class NotAcceptable(wexc.NotAcceptable, ApiException):
+    """Not acceptable"""
+
+
+class RequestTimeout(wexc.RequestTimeout, ApiException):
+    """Request timeout"""
+
+
+class Conflict(wexc.Conflict, ApiException):
+    """Conflict"""
+
+
+class Gone(wexc.Gone, ApiException):
+    """Resource gone"""
+
+
+class LengthRequired(wexc.LengthRequired, ApiException):
+    """Length required"""
+
+
+class PreconditionFailed(wexc.PreconditionFailed, ApiException):
+    """Etag required and wrong ETag provided"""
+
+
+class RequestEntityTooLarge(wexc.RequestEntityTooLarge, ApiException):
+    """Request entity too large"""
+
+
+class RequestURITooLarge(wexc.RequestURITooLarge, ApiException):
+    """Request URI too large"""
+
+
+class UnsupportedMediaType(wexc.UnsupportedMediaType, ApiException):
+    """Unsupported media type"""
+
+
+class RequestedRangeNotSatisfiable(wexc.RequestedRangeNotSatisfiable, ApiException):
+    """Requested range not satisfiable"""
+
+
+class ExpectationFailed(wexc.ExpectationFailed, ApiException):
+    """Expectation failed"""
+
+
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -157,4 +157,4 @@ cython_debug/
 #  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
-#.idea/
+.idea/
```
lafrech: Doesn't this belong to your [global .gitignore](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/)?

src/flask_smorest/exceptions.py
```diff
@@ -7,30 +7,166 @@ class FlaskSmorestError(Exception):
...
+# API exceptions
+class ApiException(wexc.HTTPException, FlaskSmorestError):
+    """A generic API error."""
+
+    @classmethod
+    def get_exception_by_code(cls, code):
+        for exception in cls.__subclasses__():
+            if exception.code == code:
+                return exception
+        return InternalServerError
```
lafrech, line 28: I think I'd return a `ValueError` here as `code` is wrong (it will result in an internal error anyway).

src/flask_smorest/exceptions.py
```diff
@@ -7,30 +7,166 @@ class FlaskSmorestError(Exception):
...
+class CurrentApiNotAvailableError(FlaskSmorestError):
+    """`current_api` not available"""
 
-    Exception created to compensate for a lack in Werkzeug (and Flask)
-    """
 
-    code = 304
-    description = "Resource not modified since last request."
+# API exceptions
+class ApiException(wexc.HTTPException, FlaskSmorestError):
```
lafrech, line 20: Wondering it it needs to inherit `HTTPException`. Perhaps. No objection, just wondering.

src/flask_smorest/exceptions.py
```diff
@@ -7,30 +7,166 @@ class FlaskSmorestError(Exception):
...
+
+
+def abort(http_status_code, exc=None, **kwargs):
+    try:
+        raise ApiException.get_exception_by_code(http_status_code)
+    except ApiException as err:
+        err.data = kwargs
+        err.exc = exc
+        raise err
+    except Exception as err:
```
lafrech, line 171: What's the point of catching any exception to re-raise it?

src/flask_smorest/exceptions.py
```diff
@@ -7,30 +7,166 @@ class FlaskSmorestError(Exception):
...
+    """
+
+    code = 304
+    description = "Resource not modified since last request."
+
+
+def abort(http_status_code, exc=None, **kwargs):
+    try:
+        raise ApiException.get_exception_by_code(http_status_code)
+    except ApiException as err:
```
lafrech, line 167: Do we have to raise and catch? Can't we tweak the instance before raising?

src/flask_smorest/exceptions.py
```diff
@@ -7,30 +7,166 @@ class FlaskSmorestError(Exception):
...
+    """Resource was not modified (Etag is unchanged)
+
+    Exception created to compensate for a lack in Werkzeug (and Flask)
+    """
+
+    code = 304
+    description = "Resource not modified since last request."
+
+
+def abort(http_status_code, exc=None, **kwargs):
```
lafrech, line 164: I like to keep only exceptions in this file.

We may want to put this somewhere else. It could be in `error_handler.py`.

lafrech (commented):
Thanks for your effort on this.

I'm not really a fan of the idea of redefining all exceptions manually, although I reckon they shouldn't change much upstream. I'd find it nice if we could create those exceptions automatically, then add the already existing changes.

Regarding raising `ApiException`, ideally it should be treated like an `api_abort`. We could decide the exception itself is private API and users must use `api_abort`, but I'd rather not.

src/flask_smorest/exceptions.py
TGoddessana:
```python
class NotModified(wexc.HTTPException, FlaskSmorestError):
    """Resource was not modified (Etag is unchanged)

    Exception created to compensate for a lack in Werkzeug (and Flask)
    """

    code = 304
    description = "Resource not modified since last request."


class PreconditionRequired(wexc.PreconditionRequired, FlaskSmorestError):
    """Etag required but missing"""

    # Overriding description as we don't provide If-Unmodified-Since
    description = 'This request is required to be conditional; try using "If-Match".'


class PreconditionFailed(wexc.PreconditionFailed, FlaskSmorestError):
    """Etag required and wrong ETag provided"""

```

This is part of original exceptions.py for flask_smorest. As you can see here, flask_smorest inherits and overrides a few HttpExceptions, so I wanted to make sure that I was calling them "http API errors", so I overridden them. (This is good, in my opinion, because it makes it clear what are API errors and what are internal errors that are not).

.gitignore
TGoddessana: I actually didn't realize this was possible, so I reverted that commit.🫠

src/flask_smorest/exceptions.py
TGoddessana:
```python
# API exceptions
class ApiException(wexc.HTTPException, FlaskSmorestError):
    """A generic API error."""

    @classmethod
    def get_exception_by_code(cls, code):
        for exception in cls.__subclasses__():
            if exception.code == code:
                return exception
        return ValueError(f"Not a standard HTTP status code: {code}")
```

I think this error message is appropriate, but if you have another one, please let me know.. :)

src/flask_smorest/exceptions.py
TGoddessana: I'll fix the location of that code, I think it's a good idea to fix it as you say

src/flask_smorest/exceptions.py
TGoddessana:
```python
class PreconditionRequired(wexc.PreconditionRequired, ApiException):
    """Etag required but missing"""

    # Overriding description as we don't provide If-Unmodified-Since
    description = 'This request is required to be conditional; try using "If-Match".'


class NotModified(ApiException):
    """Resource was not modified (Etag is unchanged)

    Exception created to compensate for a lack in Werkzeug (and Flask)
    """

    code = 304
    description = "Resource not modified since last request."


for code, exc in wexc.default_exceptions.items():
    api_exc = type(exc.__name__, (exc, ApiException), {})
    setattr(module, exc.__name__, api_exc)

```

As you said, with this approach, we could subclass all Werkzeug errors to our exception class. 

image

I'm a little curious here. Should we also allow users to import and use syntax like `from flask_smorest.exceptions import NotFound`? Because it doesn't seem to work when importing in code, not in the REPL.

image

For example, what code should a user use to get a 404 error?

```python

# 1
from flask_smorest.exceptions import ApiException
raise ApiException.get_exception_by_code(404)(...)

# 2
from flask_smorest imort abort
abort(404)

# 3
from flask_smorest.exceptions import NotFound
raise NotFound(...)

#4 
# maybe.. is this possible?
from flask_smorest.exceptions import ApiException
raise ApiException(404)
```

Would love to hear your thoughts :)

PR #654, opened 2024-06-01
Update __init__.py

kanhebei:
**The original**
    
```
   _The first method：_
    from blueprint import bp
    api = Api(app)
    api.register_bluepirnt(bp)
```
    
    _The second method_
    form blueprint import bp
    api = Api()
    # # Must call init_app before registering the API blueprint
    api.init_app(app)
    
    api.register_blueprint(bp)    
    
**After the update, there are no issues with the above two writing methods, and the following new usage methods will continue to be added**
    
    api = Api()
    
    from blueprint import bp
    api.register_blueprint(bp)
    # #Or register as a string
    api.register_blueprint('blueprint.bp')
    
    # # last
    api.init_app(app)

### DIFF
--- a/src/flask_smorest/__init__.py
+++ b/src/flask_smorest/__init__.py
@@ -1,5 +1,7 @@
 """Api extension initialization"""
 
+from typing import Union
+from werkzeug.utils import import_string
 from webargs.flaskparser import abort  # noqa
 
 from .spec import APISpecMixin
@@ -41,6 +43,7 @@ class Api(APISpecMixin, ErrorHandlerMixin):
     """
 
     def __init__(self, app=None, *, spec_kwargs=None, config_prefix=""):
+        self._routes = []
         self._app = app
         self._spec_kwargs = spec_kwargs or {}
         self.config_prefix = normalize_config_prefix(config_prefix)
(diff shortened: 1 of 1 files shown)


lafrech:
I get your point.

This is not strictly needed. See for instance in this app how we register blueprints after initializing `api`: https://github.com/BEMServer/bemserver-api/blob/main/src/bemserver_api/__init__.py. But I don't mind adding this.

Regarding importing as string, I don't mind either. Since the work is done in werkzeug, it is not too much API surface added.

Don't bother adding type hints. There is no CI check for them yet. However, anyone is welcome to add type hints to the code base in another PR.

I'd move the part that does the registration to a dedicated `_register_blueprint` method that would be called at `__init__` (no need to call `import_string` and check `self._app` at this stage). Not a performance issue, but to make the intent clear.

pat-jpnk:
> Don't bother adding type hints. There is no CI check for them yet. However, anyone is welcome to add type hints to the code base in another PR.

You mean it would make sense to add it once the whole code base is covered?

lafrech:
Le Tue, 16 Jul 2024 14:48:02 -0700,
pat-jpnk ***@***.***> a écrit :

> > Don't bother adding type hints. There is no CI check for them yet.
> > However, anyone is welcome to add type hints to the code base in
> > another PR.  
> 
> You mean it would make sense to add it once the whole code base is
> covered?

I mean don't bother adding type hints to a function in a PR until a PR
adds a significant amount of type hints and a call to mypy in CI.

-- 
Jérôme

PR #682, opened 2024-08-20
Reduce pre-commit.ci updates to monthly

sloria:
To reduce PR noise


lafrech:
Thanks.

And congratulations for your first commit in this repo!

PR #744, opened 2025-01-08
Prevent duplicate builds

sloria:
just noticed the dupe builds while peeking at PRs:

CleanShot 2025-01-07 at 23 47 19@2x

this will prevent those


lafrech:
This is a remain of shy old times where I wanted to have a silent feedback about my code before opening a PR and failing CI in public but this doesn't make much sense and is a waste of resources indeed.

PR #799, opened 2025-09-03
fix(async views): use ensure_sync in view decorators

### LINKED ISSUE #798: decorators not calling current_app.ensure_sync -> incompatible with async views
Hi, I want to gradually migrate an app to async, sadly I can't do step 1 which is to make my views async since flask_smorest doesn't support it.
See the flask docs on the mater: https://flask.palletsprojects.com/en/stable/async-await/#extensions

awoimbee:
Closes #798

~~Sorry I didn't add new tests, I didn't know where to add them since the existing tests are more about OpenAPI schema than testing the response.~~
Tests added !

### DIFF
--- a/src/flask_smorest/spec/plugins.py
+++ b/src/flask_smorest/spec/plugins.py
@@ -7,7 +7,6 @@ import werkzeug.routing
 
 from apispec import BasePlugin
 
-# from flask-restplus
 RE_URL = re.compile(r"<(?:[^:<>]+:)?([^<>]+)>")
 
 
--- a/src/flask_smorest/spec/plugins.py
+++ b/src/flask_smorest/spec/plugins.py
@@ -7,6 +7,7 @@ import werkzeug.routing
 
 from apispec import BasePlugin
 
+# from flask-restplus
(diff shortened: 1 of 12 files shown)


src/flask_smorest/spec/plugins.py
```diff
@@ -7,7 +7,6 @@
 
 from apispec import BasePlugin
 
-# from flask-restplus
```
lafrech: ?

lafrech (commented):
Thanks for contributing. This makes sense.

Regarding the tests, I think I'd add an async test along each decorator's test suite, but a new test file with all async test would be fine too.

src/flask_smorest/spec/plugins.py
awoimbee: Oops, reverted

lafrech (changes requested):
Thanks.
Looks like there's a test missing for `arguments`, but otherwise, looks good to me!

lafrech:
Thanks.

This is in 0.46.2.

